PR #22, opened 2019-10-09
:art: Add fetching benchmark datasets task

rinx:


kpango:
Could you please delete hdf5 file if checksum is invalid.
and @kou-m and me are considering about changing dataset directory structure please wait a while.

PR #68, opened 2019-11-28
:children_crossing: gRPC status codes

rinx:


rinx:
These components/methods are not implemented yet.
- gateway
- discoverer
- backup-manager
- meta: multiGet/multiRemove etc...

PR #87, opened 2019-12-11
[WIP] added test case for backoff

hlts2:
I added test code of internal/backoff.
I have confirmed that go test is passed.


internal/backoff/backoff_test.go
```diff
@@ -19,18 +19,183 @@ package backoff
...
+				backoffTimeLimit: 5 * time.Minute,
+				maxDuration:      float64(time.Hour),
+				jitterLimit:      float64(time.Minute),
+				backoffFactor:    1.5,
+				maxRetryCount:    50,
+				errLog:           true,
+				durationLimit:    float64(time.Hour) / 1.5,
+			},
+			checkFunc: func(got *backoff, want *backoff) error {
+				got.jittedInitialDuration, want.jittedInitialDuration = 1, 1
```
hlts2: Because `jittedInitialDuration` is generated from random numbers, it is difficult to compare, so ignore it

PR #96, opened 2019-12-13
fix manager backup

rinx:


rinx:
@kpango I fixed the problem using 'on duplicate key update' with auto_increment on not primary key column.
Now, the status of this PR is ready to be reviewed.

PR #376, opened 2020-05-14
♻️ Fix compressor readiness shutdown_duration / Fix cassandra …

rinx:
…serial_consistency value

### Description:

I fixed several fields in values.yaml.

### Related Issue:

nothing

nothing

### Environment:

- Golang Version: 1.14
- Docker Version: 19.03.5
- Kubernetes Version: 1.17.3
- NGT Version: 1.9.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [X] Refactoring [type/refactoring]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/charts/vald/values.yaml
+++ b/charts/vald/values.yaml
@@ -1019,7 +1019,10 @@ compressor:
     healths:
       liveness:
         enabled: false
-      readiness: {}
+      readiness:
+        server:
+          http:
+            shutdown_duration: 1m
     metrics:
       pprof: {}
       prometheus: {}
@@ -1312,7 +1315,7 @@ backupManager:
       # backupManager.cassandra.config.consistency -- consistency type
       consistency:  quorum
       # backupManager.cassandra.config.serial_consistency -- read consistency type
-      serial_consistency:  local_serial
+      serial_consistency:  localserial
       # backupManager.cassandra.config.username -- cassandra username
       username: root
       # backupManager.cassandra.config.password -- cassandra password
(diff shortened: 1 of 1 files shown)


charts/vald/values.yaml
```diff
@@ -1019,7 +1019,10 @@ compressor:
     healths:
       liveness:
         enabled: false
-      readiness: {}
+      readiness:
+        server:
+          http:
+            shutdown_duration: 1m
```
kpango, line 1025: I think readiness should shutdown immediately. this setting increase duration until disconnect from kubernetes service dns.

kpango: could you please describe more about why we need 1minutes duration for shutting down the readiness server?

rinx: hmm. which field affects to make compressor pods alive longer without increasing duration until disconnect?

kpango: Uh, I see. readiness should shutdown immediately so it should be 0s and for liveness, it affects pod to keep them alive.
but I know, we don't use liveness probe for agent and compressor, hmm.... 🤔

rinx: ah ha, okay understood.
we have now a backup strategy in compressor post-stop, it may be okay to be terminated suddenly caused by unexpected fails of liveness servers. :thinking:

kpango: I think so, maybe we need another phase for shutting down the process in internal server

kpango: if we can set post pre process for each server, It would be useful

rinx: i see. for now, i'm going to enable liveness and set readiness shutdown_duration to be zero. thanks

kpango: sound good to me.
please set liveness shutdown duration over 1minutes.
and we need to think about pod disruption budget

rinx: revised. please check it :smile:

PR #460, opened 2020-06-10
✨ Add options for AWS client

rinx:

### Description:

- add options for AWS client of Agent-sidecar
    - add disableSSL option
    - and another options in [aws.Config](https://docs.aws.amazon.com/sdk-for-go/api/aws/#Config) struct.

### Related Issue:

nothing.

nothing.

### Environment:

- Go Version: 1.14.3
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.11.5

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


internal/db/storage/blob/s3/option.go
```diff
@@ -59,3 +59,15 @@ func WithMaxPartSize(size int64) Option {
 		}
 	}
 }
+
+func WithMaxPartSizeKB(kb int) Option {
```
kpango, line 63: in this section, I think it's better to define measurement const like below
```
const (
    _ = iota
    KB int = 1 << (10 * iota)
    MB
    GB
    TB
)
```

kpango: also, it's better to define by the string (10kb, 4mb, 5gb, etc...) and parse it in the option func.

rinx: agree. i'll do it later.

internal/unit/unit.go
```diff
@@ -0,0 +1,36 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package unit
+
+import (
+	"code.cloudfoundry.org/bytefmt"
```
kpango, line 20: what a nice package!!!

internal/db/storage/blob/s3/session/session.go
```diff
@@ -68,5 +69,7 @@ func (s *sess) Session() (*session.Session, error) {
 		cfg = cfg.WithCredentials(creds)
 	}
 
+	cfg = cfg.WithDisableSSL(!s.enableSSL)
```
kpango, line 72: oh... I didn't know about this, hmm how about this
```suggestion
	if !s.enableSSL {
		cfg = cfg.WithDisableSSL(true)
	}
```

rinx: There are several cases using "WithDisableXXX" functions in AWS SDK. :fearful:
so do you think it is better to write like it for every cases?

kpango: I think aws SDK's all feature is enabled by default, that's why they designed like that, but our concept is disabled by default because Go's boolean value is false by default. 
I think it's better to write like it for every case but I understand it's boring.

rinx: okay. i'll do like that.

AWS SDK also has several "WithEnableXXX" functions. :rofl:
maybe they are thinking that all options should be "false" by default.

kpango: okay, our final decision is we should use EnableXXX and don't use implicit default enable configuration.

PR #481, opened 2020-06-16
Added test case for `internal/errorgroup`

hlts2:
### Description:

I added a test case for `internal/errorgroup`.

### Environment:

- Go Version: 1.14.3
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.11.5

### Types of changes:

- [ ] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/errgroup/group_test.go
+++ b/internal/errgroup/group_test.go
@@ -255,13 +255,9 @@ func TestGo(t *testing.T) {
 	type args struct {
 		f func() error
 	}
-	type generator struct {
-		do func() *group
-	}
 	type test struct {
 		name       string
 		args       args
-		generator  generator
 		checkFunc  func() error
 		beforeFunc func(args)
 		afterFunc  func(args)
@@ -271,15 +267,37 @@ func TestGo(t *testing.T) {
 	}
 	tests := []test{
 		func() test {
+			var calledCnt int32
+
+			var enableLimitation atomic.Value
+			enableLimitation.Store(false)
+
+			g := &group{
+				enableLimitation: enableLimitation,
+			}
+
 			return test{
 				name: "instance.Go is called when instance is not nil",
-				args: args{},
-				generator: generator{
-					do: func() *group {
-						return new(group)
+				args: args{
+					f: func() error {
+						atomic.AddInt32(&calledCnt, 1)
+						return nil
 					},
 				},
-				checkFunc: defaultCheckFunc,
+				beforeFunc: func(args) {
+					instance = g
+				},
+				checkFunc: func() error {
+					if err := g.Wait(); err != nil {
+						return err
+					}
+
+					if got, want := int(atomic.LoadInt32(&calledCnt)), 1; got != want {
+						return errors.Errorf("calledCnt = %v, want: %v", got, want)
+					}
+					return nil
+
+				},
 			}
 		}(),
 	}
--- a/internal/errgroup/group_test.go
+++ b/internal/errgroup/group_test.go
@@ -46,7 +46,8 @@ func TestNew(t *testing.T) {
 		afterFunc  func(args)
 	}
 	defaultCheckFunc := func(w want, got Group, got1 context.Context) error {
-		if !reflect.DeepEqual(got, w.want) {
+		if got, want := got.(*group), w.want.(*group); !reflect.DeepEqual(got.emap, want.emap) &&
+			!reflect.DeepEqual(got.enableLimitation, want.enableLimitation) {
 			return errors.Errorf("got = %v, want %v", got, w.want)
 		}
 		if !reflect.DeepEqual(got1, w.want1) {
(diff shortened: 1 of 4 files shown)


internal/errgroup/group_test.go
```diff
@@ -46,7 +46,8 @@ func TestNew(t *testing.T) {
 		afterFunc  func(args)
 	}
 	defaultCheckFunc := func(w want, got Group, got1 context.Context) error {
-		if !reflect.DeepEqual(got, w.want) {
+		if got, want := got.(*group), w.want.(*group); !reflect.DeepEqual(got.emap, want.emap) &&
+			!reflect.DeepEqual(got.enableLimitation, want.enableLimitation) {
 			return errors.Errorf("got = %v, want %v", got, w.want)
```
kevindiu, line 51: do we need to check if `cancel` field is not nil?

internal/errgroup/group_test.go
```diff
@@ -307,86 +328,78 @@ func Test_group_Limitation(t *testing.T) {
...
-		           emap: nil,
-		           errs: nil,
-		           err: nil,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		func() test {
```
kevindiu, line 357: how about rewrite it like this? I think it is easier to read, what do you think?
```golang
357         {
358             name: "set disable when limit is 0",
359             args: args{
360                 limit: 0,
361             },
362             fields: fields{
363                 limitation:       make(chan struct{}),
364                 enableLimitation: atomic.Value{},
365             },
366             want: want{
367                 want: &group{
368                     enableLimitation: func() atomic.Value {
369                         var el atomic.Value
370                         el.Store(false)
371                         return el
372                     }(),
373                 },
374             },
375         },
```

internal/errgroup/group_test.go
```diff
@@ -307,86 +328,78 @@ func Test_group_Limitation(t *testing.T) {
...
+				},
+				want: want{
+					want: &group{
+						enableLimitation: el,
+					},
+				},
+			}
+		}(),
+
+		func() test {
```
kevindiu, line 378: same as before

internal/errgroup/group_test.go
```diff
@@ -307,86 +328,78 @@ func Test_group_Limitation(t *testing.T) {
...
-		   }(),
-		*/
+		func() test {
+			var el atomic.Value
+			el.Store(false)
+
+			return test{
+				name: "set disable when limit is 0",
+				args: args{
+					limit: -1,
```
kevindiu, line 364: it should be 0?

internal/errgroup/group_test.go
```diff
@@ -511,102 +586,58 @@ func Test_group_Go(t *testing.T) {
 			if test.checkFunc == nil {
 				test.checkFunc = defaultCheckFunc
 			}
-			g := &group{
```
kevindiu, line 514: I think it is better than `generator`, even it has lint error. what do you think?

kevindiu, line 514: I think it is better than `generator`, what do you think? (even it has lint error)

hlts2: I understand your opinion.
but I think we do not ignore lint error.
in this case, copying the mutex causes an error.
what do you think.

kevindiu: maybe just remove line 521?

internal/errgroup/group_test.go
```diff
@@ -421,87 +426,157 @@ func Test_group_Go(t *testing.T) {
...
+
+			var enableLimitation atomic.Value
+			enableLimitation.Store(true)
+
+			egctx, cancel := context.WithCancel(context.Background())
+			cancel()
+
+			g := &group{
+				egctx:            egctx,
+				limitation:       make(chan struct{}),
```
kevindiu, line 455: It should be 
```suggestion
				limitation:       make(chan struct{}, 0),
```

internal/errgroup/group_test.go
```diff
@@ -421,87 +426,157 @@ func Test_group_Go(t *testing.T) {
...
+				enableLimitation: enableLimitation,
+			}
+
+			return test{
+				name: "f is not called when reached limit and cancel g.egctx",
```
kevindiu, line 460: I think we also need to test the limitation function (for example the limit is 3, and 5 `Go(f func error)` is executed and only 3 can be executed at first, any after 1 is completed the 4th func will execute)

internal/errgroup/group_test.go
```diff
@@ -421,87 +426,157 @@ func Test_group_Go(t *testing.T) {
...
+					cancel()
+					atomic.AddInt32(&cancelCalledCnt, 1)
+				},
+				emap:             make(map[string]struct{}),
+				limitation:       make(chan struct{}, 1),
+				enableLimitation: enableLimitation,
+			}
+
+			return test{
+				name: "f is called and f returns error",
```
kevindiu, line 547: do we need to check multiple error is returned?

internal/errgroup/group_test.go
hlts2: ok.
but if the test use mutex or sync.once variable, I use a generator.
what do you think?

internal/errgroup/group_test.go
```diff
@@ -729,49 +738,34 @@ func Test_group_Wait(t *testing.T) {
...
-		           emap: nil,
-		           errs: nil,
-		           err: nil,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
```
kevindiu, line 741: do we need some test case to block the current `Go(func())` call and the `Wait()` should actually wait for all functions return?

internal/errgroup/group_test.go
hlts2: I don't think you need to write 0 explicitly.

internal/errgroup/group_test.go
kevindiu: using generator is the same. the difference of `generator` is that it is missing this statement `mu:  test.fields.mu,` , so the lint wouldn't report a error.

internal/errgroup/group_test.go
kevindiu: One is Buffered channel and one is unbuffered channel, if you think using unbuffered one is okay then it is fine for me;
https://medium.com/a-journey-with-go/go-buffered-and-unbuffered-channels-29a107c00268

internal/errgroup/group_test.go
hlts2: :memo:
we may use generator to avoid coping `waitGroup` & `sync.Mutex`

so I close this comment.:pray:

PR #526, opened 2020-06-29
Add test for `internal/file/watch`

hlts2:

### Description:

I added test case for `internal/file/watch`

### Environment:

- Go Version: 1.14.3
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.11.5

### Types of changes:

- [ ] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/file/watch/option_test.go
+++ b/internal/file/watch/option_test.go
@@ -26,13 +26,6 @@ import (
 	"go.uber.org/goleak"
 )
 
-var (
-	// Goroutine leak is detected by `fastime`, but it should be ignored in the test because it is an external package.
-	goleakIgnoreOptions = []goleak.Option{
-		goleak.IgnoreTopFunction("github.com/kpango/fastime.(*Fastime).StartTimerD.func1"),
-	}
-)
-
 func TestWithErrGroup(t *testing.T) {
 	type T = watch
 	type args struct {
--- a/internal/file/watch/watch.go
+++ b/internal/file/watch/watch.go
@@ -28,6 +28,7 @@ import (
 	"github.com/vdaas/vald/internal/safety"
 )
 
+// Watcher is an interface that represents a file monitor.
 type Watcher interface {
 	Start(ctx context.Context) (<-chan error, error)
 	Add(dirs ...string) error
@@ -49,6 +50,7 @@ type watch struct {
 	onError  func(ctx context.Context, err error) error
 }
 
+// New returns Watcher implementation.
 func New(opts ...Option) (Watcher, error) {
 	w := new(watch)
 	for _, opt := range append(defaultOpts, opts...) {
@@ -93,6 +95,8 @@ func (w *watch) init() (*watch, error) {
 	return w, nil
 }
 
+// Start starts watching all named files or directories. If an error occurs, returns the error.
+// And performs the processing corresponding to the file change event, and returns an error via channel if an error occurs in them.
 func (w *watch) Start(ctx context.Context) (<-chan error, error) {
 	ech := make(chan error, 10)
 	w.eg.Go(safety.RecoverFunc(func() (err error) {
@@ -158,6 +162,7 @@ func (w *watch) Start(ctx context.Context) (<-chan error, error) {
 	return ech, nil
 }
 
+// Add starts watching all named files or directories. If an error occurs, returns the error.
 func (w *watch) Add(dirs ...string) (err error) {
 	w.mu.Lock()
 	defer w.mu.Unlock()
@@ -173,6 +178,7 @@ func (w *watch) Add(dirs ...string) (err error) {
 	return nil
 }
 
+// Remove stops watching all named files or directories. If an error occurs, returns the error.
 func (w *watch) Remove(dirs ...string) (err error) {
 	w.mu.Lock()
 	defer w.mu.Unlock()
@@ -188,6 +194,7 @@ func (w *watch) Remove(dirs ...string) (err error) {
 	return nil
 }
 
+// Stop stops watching all named files or directories. If an error occurs, returns the error.
 func (w *watch) Stop(ctx context.Context) (err error) {
 	w.mu.Lock()
 	defer w.mu.Unlock()
--- a/internal/file/watch/watch.go
+++ b/internal/file/watch/watch.go
@@ -145,7 +145,10 @@ func (w *watch) Start(ctx context.Context) (<-chan error, error) {
 			case err, ok = <-w.w.Errors:
 			}
 			if !ok {
-				w, err = w.init()
+				iw, err := w.init()
+				if err == nil {
+					w = iw
+				}
 			}
 			if err != nil {
 				handleErr(ctx, err)
@@ -159,12 +162,12 @@ func (w *watch) Add(dirs ...string) (err error) {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 	for _, dir := range dirs {
-		w.dirs[dir] = struct{}{}
 		if w.w != nil {
 			err = w.w.Add(dir)
 			if err != nil {
 				return err
 			}
+			w.dirs[dir] = struct{}{}
 		}
 	}
 	return nil
--- a/internal/file/watch/watch.go
+++ b/internal/file/watch/watch.go
@@ -71,12 +71,12 @@ func (w *watch) init() (*watch, error) {
 		return nil, err
 	}
 	w.mu.RLock()
-	dirs := make(map[string]struct{}, len(w.dirs))
-	for name, val := range w.dirs {
-		dirs[name] = val
+	dirs := make([]string, 0, len(w.dirs))
+	for d := range w.dirs {
+		dirs = append(dirs, d)
 	}
 	w.mu.RUnlock()
-	for dir := range dirs {
+	for _, dir := range dirs {
 		log.Debug("Adding watcher target: ", dir)
 
 		err = watcher.Add(dir)
--- a/internal/file/watch/watch.go
+++ b/internal/file/watch/watch.go
@@ -71,7 +71,10 @@ func (w *watch) init() (*watch, error) {
 		return nil, err
 	}
 	w.mu.RLock()
-	dirs := w.dirs
+	dirs := make(map[string]struct{}, len(w.dirs))
+	for name, val := range w.dirs {
+		dirs[name] = val
+	}
 	w.mu.RUnlock()
 	for dir := range dirs {
 		log.Debug("Adding watcher target: ", dir)
--- a/internal/file/watch/watch_test.go
+++ b/internal/file/watch/watch_test.go
@@ -132,20 +132,18 @@ func Test_watch_init(t *testing.T) {
 	type test struct {
 		name       string
 		fields     fields
-		wantFunc   func(*testing.T) want
+		want       want
 		checkFunc  func(want, *watch, error) error
-		beforeFunc func()
+		beforeFunc func(*testing.T, *fields)
 		afterFunc  func()
 	}
 	defaultCheckFunc := func(w want, got *watch, err error) error {
 		if !errors.Is(err, w.err) {
 			return errors.Errorf("got error = %v, want %v", err, w.err)
 		}
-		// if w.want != nil {
-		// 	if !reflect.DeepEqual(got.w, w.want.w) {
-		// 		return errors.Errorf("got = %v, want %v", got, w.want)
-		// 	}
-		// }
+		if !reflect.DeepEqual(got, w.want) {
+			return errors.Errorf("got = %v, want %v", got, w.want)
+		}
 		return nil
 	}
 	tests := []test{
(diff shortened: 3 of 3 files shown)


kevindiu:
do we need author review for this PR?

hlts2:
@kevindiu 
 
> do we need author review for this PR?

I don't think it is necessary.
Because I wrote code with @kpango (pair programming).

kevindiu:
> @kevindiu
> 
> > do we need author review for this PR?
> 
> I don't think it is necessary.
> Because I wrote code with @kpango (pair programming).

OK!

kevindiu:
this error occurred...
```
% go test
# github.com/vdaas/vald/internal/file/watch [github.com/vdaas/vald/internal/file/watch.test]
./watch_test.go:36:2: goleakIgnoreOptions redeclared in this block
	previous declaration at ./option_test.go:31:2
FAIL	github.com/vdaas/vald/internal/file/watch [build failed]
```

kevindiu:
please write comments on `watch.go` :)

kevindiu:
I am not sure but do we need to lock here?
https://github.com/vdaas/vald/pull/526/files#diff-7c1a204e5f48377cbbca258e20ea87dfL72

as it only assign the local pointer variable and should not have any affect on locking

internal/file/watch/watch_test.go
```diff
@@ -139,58 +150,48 @@ func Test_watch_init(t *testing.T) {
...
+				},
+			},
+			checkFunc: func(w want, got *watch, err error) error {
+				if !errors.Is(err, w.err) {
+					return errors.Errorf("got error = %v, want %v", err, w.err)
+				}
+				if got == nil {
+					return errors.New("got is nil")
+				}
+				if got.w == nil {
```
kevindiu, line 179: can we check the internal watching files of got.w?

internal/file/watch/watch_test.go
```diff
@@ -139,58 +150,48 @@ func Test_watch_init(t *testing.T) {
...
+			},
+			want: want{
+				err: syscall.Errno(0x2),
+			},
+		},
+
+		{
+			name: "returns (*watch, nil) when no error occurs internally",
+			fields: fields{
+				dirs: map[string]struct{}{
```
kevindiu, line 168: please add multiple files test (with all exists file test and some non-exists file)

hlts2:
@kevindiu 
> I am not sure but do we need to lock here?

I think this logic is necessary. Because dirs can change dynamically.

kevindiu:
> @kevindiu
> 
> > I am not sure but do we need to lock here?
> 
> I think this logic is necessary. Because dirs can change dynamically.

but it just copy the pointer to the local variable `dirs` instead of coping the map to local variable.
If we do locking I think we need to unlock it on line 83?

internal/file/watch/watch_test.go
```diff
@@ -354,118 +619,107 @@ func Test_watch_Add(t *testing.T) {
...
+			fields: fields{
+				dirs: make(map[string]struct{}),
+			},
+			want: want{
+				err: syscall.Errno(0x2),
+				want: &watch{
+					dirs: make(map[string]struct{}),
+				},
+			},
+		},
```
kevindiu, line 699: Do you think we need to test if we add the same file name?
I am not sure but I think there is a potential bug that if we add the same file with different path, the same will be added twice.

for example  adding "./watch.go" and "../watch/watch.go"

internal/file/watch/watch_test.go
```diff
@@ -478,121 +732,123 @@ func Test_watch_Remove(t *testing.T) {
...
+			want: want{
+				want: &watch{
+					dirs: map[string]struct{}{},
+				},
+				err: nil,
+			},
+		},
+
+		{
+			name: "returns error when w.w.Remove returns non-exist inotify error",
```
kevindiu, line 805:
```suggestion
			name: "returns  non-exist error when w.w.Remove remove non-exists file",
```

internal/file/watch/watch_test.go
```diff
@@ -354,118 +619,107 @@ func Test_watch_Add(t *testing.T) {
...
+				want: &watch{
+					dirs: map[string]struct{}{
+						"./watch.go":  struct{}{},
+						"./option.go": struct{}{},
+					},
+				},
+			},
+		},
+
+		{
```
kevindiu, line 683: do we need test case for adding folders?

internal/file/watch/watch_test.go
```diff
@@ -478,121 +732,123 @@ func Test_watch_Remove(t *testing.T) {
...
+					"watch_test.go": struct{}{},
+				},
+			},
+			want: want{
+				want: &watch{
+					dirs: map[string]struct{}{},
+				},
+				err: nil,
+			},
+		},
```
kevindiu, line 802: do we need removing folder test case?

kevindiu: also we need partial remove test case?

internal/file/watch/watch_test.go
```diff
@@ -478,121 +732,123 @@ func Test_watch_Remove(t *testing.T) {
...
+
+		if got, want := len(got.dirs), len(w.want.dirs); got != want {
+			return errors.Errorf("dirs length = %d, want %d", got, want)
+		}
+
+		for name := range w.want.dirs {
+			if _, ok := got.dirs[name]; !ok {
+				return errors.Errorf("dirs %s is not exists", name)
+			}
+		}
```
kevindiu, line 764: can we verify the internal `*fsnotify.Watcher` changes?

internal/file/watch/watch_test.go
```diff
@@ -354,118 +619,107 @@ func Test_watch_Add(t *testing.T) {
...
 			return errors.Errorf("got error = %v, want %v", err, w.err)
 		}
+		if got, want := len(got.dirs), len(w.want.dirs); got != want {
+			return errors.Errorf("dirs length = %d, want %d", got, want)
+		}
+		for name := range w.want.dirs {
+			if _, ok := got.dirs[name]; !ok {
+				return errors.Errorf("dirs %s is not exists", name)
+			}
+		}
```
kevindiu, line 649: can we verify the internal changes of `*fsnotify.Watcher`?

internal/file/watch/watch_test.go
```diff
@@ -602,95 +858,106 @@ func Test_watch_Stop(t *testing.T) {
...
-		           onChmod: nil,
-		           onError: nil,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "returns nil when w.Remove returns nil",
```
kevindiu, line 894:
```suggestion
			name: "returns nil when stop success",
```

internal/file/watch/watch_test.go
```diff
@@ -602,95 +858,106 @@ func Test_watch_Stop(t *testing.T) {
...
+			want: want{
+				want: &watch{
+					dirs: make(map[string]struct{}),
+				},
+				err: nil,
+			},
+		},
+
+		{
+			name: "returns error when w.Remove returns non-exist inotify error",
```
kevindiu, line 929:
```suggestion
			name: "returns non-exist error when the file not exists",
```

internal/file/watch/watch_test.go
hlts2: I don't think we have a responsibility to know the details beyond.
If it fails, an error will be returned, so it may be enough to make that judgment.
what do you think?

internal/file/watch/watch_test.go
hlts2: https://github.com/vdaas/vald/pull/526#discussion_r450736130

internal/file/watch/watch_test.go
kevindiu: I am not sure.
The `w.Add()` function will starting watching the file, and we can't verify if the file is being watching or not... :(
I think the implementation of the library is bad :(

internal/file/watch/watch_test.go
```diff
@@ -255,68 +245,342 @@ func Test_watch_Start(t *testing.T) {
...
-		if !reflect.DeepEqual(got, w.want) {
-			return errors.Errorf("got = %v, want %v", got, w.want)
+
+		if !errors.Is(<-got, <-w.want) {
+			return errors.Errorf("errCh got = %v, want %v", got, w.want)
 		}
+
 		return nil
 	}
 	tests := []test{
```
kevindiu, line 255: please fix the test case name :)

internal/file/watch/watch_test.go
```diff
@@ -255,68 +245,342 @@ func Test_watch_Start(t *testing.T) {
...
+					want: func() chan error {
+						ch := make(chan error, 1)
+						ch <- errors.New("err")
+						close(ch)
+						return ch
+					}(),
+					err: nil,
+				},
+			}
+		}(),
```
kevindiu, line 578: thinking we need a test case of 
1. keep watching the file change for multiple times
2. (not sure) create a test file and actually change the file.

I am thinking this test case is not effective as it doesn't really watching the file, It just creating a fake event and test the behavior of that fake event.

kevindiu: about the first case, what if the file change for multiple times with the same event?

internal/file/watch/watch.go
```diff
@@ -69,7 +71,10 @@ func (w *watch) init() (*watch, error) {
 		return nil, err
 	}
 	w.mu.RLock()
-	dirs := w.dirs
+	dirs := make(map[string]struct{}, len(w.dirs))
```
kevindiu, line 74: I think slice is better than map in this case.

internal/file/watch/watch.go
```diff
@@ -69,7 +71,10 @@ func (w *watch) init() (*watch, error) {
...
+	for name, val := range w.dirs {
+		dirs[name] = val
+	}
 	w.mu.RUnlock()
 	for dir := range dirs {
```
kevindiu, line 79:
```suggestion
	dirs := make([]string, 0, len(w.dirs))
	for _, d := range w.dirs {
		dirs = append(dirs, d)
	}
	w.mu.RUnlock()
	for _, dir := range dirs {
```

internal/file/watch/option_test.go
```diff
@@ -26,13 +26,6 @@ import (
 	"go.uber.org/goleak"
 )
 
-var (
-	// Goroutine leak is detected by `fastime`, but it should be ignored in the test because it is an external package.
-	goleakIgnoreOptions = []goleak.Option{
-		goleak.IgnoreTopFunction("github.com/kpango/fastime.(*Fastime).StartTimerD.func1"),
-	}
-)
-
```
vankichi, line 35: Does this update not affect test, e.g. https://github.com/vdaas/vald/pull/526/files#diff-9709653ad164a1d234b584cfc3aee17aR80 ?

hlts2: I think this update not affect.
There are multiple ignores and it was complicated, so watch_test.go Moved to
https://github.com/vdaas/vald/pull/526/files#diff-067a37d43f4012cebe5f5931eda3a8aaR36-R43

vankichi: this link won't work 😓

internal/file/watch/watch.go
```diff
@@ -69,9 +71,12 @@ func (w *watch) init() (*watch, error) {
 		return nil, err
 	}
 	w.mu.RLock()
-	dirs := w.dirs
+	dirs := make([]string, 0, len(w.dirs))
+	for d := range w.dirs {
+		dirs = append(dirs, d)
+	}
```
vankichi, line 77: let me tell about why, please?

internal/file/watch/watch.go
```diff
@@ -145,7 +152,10 @@ func (w *watch) Start(ctx context.Context) (<-chan error, error) {
 			case err, ok = <-w.w.Errors:
 			}
 			if !ok {
-				w, err = w.init()
+				iw, err := w.init()
+				if err == nil {
+					w = iw
+				}
```
vankichi, line 158: same. Please tell me the reason.

hlts2: sure.
if w.init returns error, w become nil.
and when roop continues, nil pointer exception will occur.

internal/file/watch/watch.go
hlts2: sure.
map type is reference type in golang.
so Even if we take the lock w.dirs, it changes when the referrer changes.
When we copy the value of w.dirs, the instantaneous state is copied, so even if the referrer changes, the copy is not affected.

internal/file/watch/option_test.go
hlts2: sorry.
Could you please read this lines??
https://github.com/vdaas/vald/blob/test/internal/file_watch/internal/file/watch/watch_test.go#L35-L42

vankichi:
@hlts2 
When you update some logics, please mention about that and author review is required or not at the PR description field.

PR #539, opened 2020-07-02
tensorflow savedmodel warmup

datelier:
Signed-off-by: datelier <57349093+datelier@users.noreply.github.com>

### Description:
tensorflow savedmodel warmup 

### Environment:

- Go Version: 1.14.3
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.11.5

### Types of changes:

- [ ] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


internal/core/converter/tensorflow/tensorflow.go
```diff
@@ -69,7 +70,17 @@ const (
 	threeDim
 )
 
-var loadFunc = tf.LoadSavedModel
+var loadFunc = func(t *tensorflow) error {
```
kpango, line 73: local func variable is not clean, please use receiver func instead of local func val

internal/core/converter/tensorflow/option_test.go
```diff
@@ -830,6 +830,104 @@ func TestWithFetches(t *testing.T) {
...
+						"test",
+					},
+				},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(tt *testing.T) {
+			defer goleak.VerifyNone(t)
```
kevindiu, line 908:
```suggestion
			defer goleak.VerifyNone(tt)
```

hlts2, line 908:
```suggestion
			defer goleak.VerifyNone(tt)
```

internal/core/converter/tensorflow/tensorflow.go
kevindiu: I think he want to override it during testing, so he defines this function as variable instead of the receiver function.

One more thing, this function is not tested ... (the coverage of this file is 86%)

kpango: exactly, that's why I recommend method & it can be replaced by Interface, please mock it using interface

internal/core/converter/tensorflow/option_test.go
```diff
@@ -482,6 +485,92 @@ func TestWithTags(t *testing.T) {
...
+			return errors.Errorf("err: %s", diff)
+		}
+		opt := cmp.Comparer(func(want, obj T) bool {
+			p1 := reflect.ValueOf(want).FieldByName("loadFunc").Pointer()
+			p2 := reflect.ValueOf(obj).FieldByName("loadFunc").Pointer()
+			return p1 == p2
+		})
+		if !cmp.Equal(w.obj, obj, opt) {
+			return errors.Errorf("got = %v, want = %v", obj, w.obj)
+		}
```
hlts2, line 521:
```suggestion
			opts := []cmp.Option{
			cmp.AllowUnexported(tensorflow{}),
			cmp.AllowUnexported(OutputSpec{}),
			cmpopts.IgnoreFields(tensorflow{}, "loadFunc"),
			cmp.Comparer(func(want, got TF) bool {
				p1 := reflect.ValueOf(want).Elem().FieldByName("loadFunc").Pointer()
				p2 := reflect.ValueOf(got).Elem().FieldByName("loadFunc").Pointer()
				return p1 == p2
			}),
		}
		if diff := cmp.Diff(w.want, got, opts...); diff != "" {
			return errors.Errorf("err: %s", diff)
		}
```

internal/core/converter/tensorflow/option.go
```diff
@@ -138,6 +152,19 @@ func WithFetches(operationNames []string, outputIndexes []int) Option {
 	}
 }
 
+// WithWarmupInputs returns Option that sets warmupInputs.
+func WithWarmupInputs(warmupInputs ...string) Option {
+	return func(t *tensorflow) {
+		if warmupInputs != nil {
+			if t.warmupInputs != nil {
+				t.warmupInputs = append(t.warmupInputs, warmupInputs...)
```
kevindiu, line 160: thinking functional options should not append inputs instead of replacing the value only.
@hlts2 what do you think?

hlts2: I understand your opinion.
but From a flexibility point of view, I would like to keep the current implementation.

PR #589, opened 2020-07-20
Add internal roundtrip test code

kevindiu:
Author review required?

### Description:

This PR implements the roundtrip test case.
This PR also fix the bug that it did not return `errors.ErrTransportRetryable` when the roundtrip success and the request is retryable.
Also this PR update the roundtrip logic that making the request terminated if the request is not retryable.

### Environment:

- Go Version: 1.14.4
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.0

### Types of changes:

- [x] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/net/http/transport/roundtrip.go
+++ b/internal/net/http/transport/roundtrip.go
@@ -51,7 +51,7 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 		return e.roundTrip(req)
 	}
 
-	var fnerr error
+	var rterr error
 	_, err = e.bo.Do(req.Context(), func() (interface{}, error) {
 		r, e := e.roundTrip(req)
 		if e != nil {
@@ -60,7 +60,7 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 				return nil, e
 			}
 			// if the error is not retryable, return nil error to terminate the backoff execution
-			fnerr = e
+			rterr = e
 			return nil, nil
 		}
 		res = r
@@ -69,8 +69,8 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 	if err != nil {
 		return nil, err
 	}
-	if fnerr != nil {
-		return nil, fnerr
+	if rterr != nil {
+		return nil, rterr
 	}
 
 	return res, nil
--- a/internal/net/http/transport/roundtrip.go
+++ b/internal/net/http/transport/roundtrip.go
@@ -51,31 +51,28 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 		return e.roundTrip(req)
 	}
 
-	var ferr error
-	r, err := e.bo.Do(req.Context(), func() (res interface{}, err error) {
-		res, err = e.roundTrip(req)
-		if err != nil {
+	var boerr error
+	_, err = e.bo.Do(req.Context(), func() (interface{}, error) {
+		r, e := e.roundTrip(req)
+		if e != nil {
 			// if the error is retryable, return the error and let backoff to retry.
-			if errors.Is(err, errors.ErrTransportRetryable) {
-				return nil, err
+			if errors.Is(e, errors.ErrTransportRetryable) {
+				return nil, e
 			}
 			// if the error is not retryable, return nil error to terminate the backoff execution
-			ferr = err
+			boerr = e
 			return nil, nil
 		}
-		return res, nil
+		res = r
+		return r, nil
 	})
 	if err != nil {
 		return nil, err
 	}
-	if ferr != nil {
-		return nil, ferr
+	if boerr != nil {
+		return nil, boerr
 	}
 
-	res, ok := r.(*http.Response)
-	if !ok {
-		return nil, errors.ErrInvalidTypeConversion(r, res)
-	}
 	return res, nil
 }
 
@@ -85,7 +82,7 @@ func (e *ert) roundTrip(req *http.Request) (res *http.Response, err error) {
 		if res != nil { // just in case we check the response as it depends on RoundTrip impl.
 			closeBody(res.Body)
 			if retryable(res) {
-				return nil, errors.ErrTransportRetryable
+				return nil, errors.Wrap(errors.ErrTransportRetryable, err.Error())
 			}
 		}
 		return nil, err
--- a/internal/net/http/transport/roundtrip.go
+++ b/internal/net/http/transport/roundtrip.go
@@ -51,11 +51,25 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 		return e.roundTrip(req)
 	}
 
-	r, err := e.bo.Do(req.Context(), func() (interface{}, error) {
-		return e.roundTrip(req)
+	var boErr error
+	r, err := e.bo.Do(req.Context(), func() (res interface{}, err error) {
+		res, err = e.roundTrip(req)
+		if err != nil {
+			// if the error is retryable, return the error and let backoff to retry.
+			if errors.Is(err, errors.ErrTransportRetryable) {
+				return nil, err
+			}
+			// if the error is not retryable, return nil error to backoff to terminate the backoff execution
+			boErr = err
+			return nil, nil
+		}
+		return res, nil
 	})
 	if err != nil {
 		return nil, err
+	}
+	if boErr != nil {
+		return nil, boErr
 	}
 
 	res, ok := r.(*http.Response)
--- a/internal/net/http/transport/roundtrip.go
+++ b/internal/net/http/transport/roundtrip.go
@@ -42,7 +42,10 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
 	return e
 }
 
-// RoundTrip round trip the HTTP request and return the response
+// RoundTrip round trip the HTTP request and return the response.
+// If backoff is not set, the default roundTrip implementation will be used.
+// The default roundTrip implementation round trip the request, and return any round trip error occurred.
+// It returns errors.ErrTransportRetryable to indicate if the request is consider as retryable.
 func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 	if e.bo == nil {
 		return e.roundTrip(req)
@@ -65,20 +68,18 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 
 func (e *ert) roundTrip(req *http.Request) (res *http.Response, err error) {
 	res, err = e.transport.RoundTrip(req)
-	if err == nil {
-		if res != nil && retryable(res) {
+	if err != nil {
+		if res != nil { // just in case we check the response as it depends on RoundTrip impl.
 			closeBody(res)
-			return nil, errors.ErrTransportRetryable
+			if retryable(res) {
+				return nil, errors.ErrTransportRetryable
+			}
 		}
-
-		return res, nil
-	}
-	if res == nil {
 		return nil, err
 	}
 
-	closeBody(res)
-	if retryable(res) {
+	if res != nil && retryable(res) {
+		closeBody(res)
 		return nil, errors.ErrTransportRetryable
 	}
 	return res, nil
--- a/internal/net/http/transport/roundtrip.go
+++ b/internal/net/http/transport/roundtrip.go
@@ -32,6 +32,7 @@ type ert struct {
 	bo        backoff.Backoff
 }
 
+// NewExpBackoff returns the backoff roundtripper implementation
 func NewExpBackoff(opts ...Option) http.RoundTripper {
 	e := new(ert)
 	for _, opt := range append(defaultOpts, opts...) {
@@ -41,6 +42,7 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
 	return e
 }
 
+// RoundTrip round trip the HTTP request and return the response
 func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 	if e.bo == nil {
 		return e.roundTrip(req)
--- a/internal/net/http/transport/roundtrip.go
+++ b/internal/net/http/transport/roundtrip.go
@@ -79,6 +79,7 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 func (e *ert) roundTrip(req *http.Request) (res *http.Response, err error) {
 	res, err = e.transport.RoundTrip(req)
 	if err != nil {
+		log.Error(err)
 		if res != nil { // just in case we check the response as it depends on RoundTrip impl.
 			closeBody(res.Body)
 			if retryableStatusCode(res.StatusCode) {
--- a/internal/net/http/transport/roundtrip.go
+++ b/internal/net/http/transport/roundtrip.go
@@ -51,7 +51,7 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 		return e.roundTrip(req)
 	}
 
-	var boerr error
+	var fnerr error
 	_, err = e.bo.Do(req.Context(), func() (interface{}, error) {
 		r, e := e.roundTrip(req)
 		if e != nil {
@@ -60,7 +60,7 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 				return nil, e
 			}
 			// if the error is not retryable, return nil error to terminate the backoff execution
-			boerr = e
+			fnerr = e
 			return nil, nil
 		}
 		res = r
@@ -69,8 +69,8 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 	if err != nil {
 		return nil, err
 	}
-	if boerr != nil {
-		return nil, boerr
+	if fnerr != nil {
+		return nil, fnerr
 	}
 
 	return res, nil
--- a/internal/net/http/transport/roundtrip.go
+++ b/internal/net/http/transport/roundtrip.go
@@ -66,19 +66,25 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 func (e *ert) roundTrip(req *http.Request) (res *http.Response, err error) {
 	res, err = e.transport.RoundTrip(req)
 	if err == nil {
+		if res != nil && retryable(res) {
+			closeBody(res)
+			return nil, errors.ErrTransportRetryable
+		}
+
 		return res, nil
 	}
 	if res == nil {
 		return nil, err
 	}
-	_, err = io.Copy(ioutil.Discard, res.Body)
-	if err != nil {
-		log.Error(err)
+
+	closeBody(res)
+	if retryable(res) {
+		return nil, errors.ErrTransportRetryable
 	}
-	err = res.Body.Close()
-	if err != nil {
-		log.Error(err)
-	}
+	return res, nil
+}
+
+func retryable(res *http.Response) bool {
 	switch res.StatusCode {
 	case http.StatusTooManyRequests,
 		http.StatusInternalServerError,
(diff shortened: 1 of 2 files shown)


internal/net/http/transport/roundtrip.go
```diff
@@ -41,50 +42,66 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
-	_, err = io.Copy(ioutil.Discard, res.Body)
+func (e *ert) roundTrip(req *http.Request) (*http.Response, error) {
+	res, err := e.transport.RoundTrip(req)
 	if err != nil {
-		log.Error(err)
+		if res != nil { // just in case we check the response as it depends on RoundTrip impl.
+			closeBody(res)
+			if retryable(res) {
+				return nil, errors.ErrTransportRetryable
+			}
```
hlts2, line 75: 🤔 
```suggestion
			}
			return res, err
```

kevindiu: I think returning `nil` instead of `res` is better 🤔
what do you think?

kpango: we already returns error at L77

hlts2: @kpango 
I'm sorry. I was reading it wrongly.

internal/net/http/transport/roundtrip.go
```diff
@@ -41,50 +42,66 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
 	return e
 }
 
-func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
+// RoundTrip round trip the HTTP request and return the response.
+// If backoff is not set, the default roundTrip implementation will be used.
+// The default roundTrip implementation round trip the request, and return any round trip error occurred.
```
rinx, line 47: badge
we don't have to mention about the behavior of the default RoundTripper. it may differ if the variable `ert.transport` set.

internal/net/http/transport/roundtrip.go
```diff
@@ -41,50 +42,66 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
+// It returns errors.ErrTransportRetryable to indicate if the request is consider as retryable.
+func (e *ert) RoundTrip(req *http.Request) (*http.Response, error) {
 	if e.bo == nil {
 		return e.roundTrip(req)
 	}
-	var r interface{}
-	r, err = e.bo.Do(req.Context(), func() (interface{}, error) {
+
+	r, err := e.bo.Do(req.Context(), func() (interface{}, error) {
 		return e.roundTrip(req)
```
rinx, line 55: badge
there's no need to check the error is `errors.ErrTransportRetryable` or not?
i think if a returned error is not `errors.ErrTransportRetryable`, it should escape the `Do` block immediately.

rinx: but i'm not sure about a correct behavior of RoundTripper. please consider about the problem above.

internal/net/http/transport/roundtrip.go
```diff
@@ -41,50 +42,66 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
+// The default roundTrip implementation round trip the request, and return any round trip error occurred.
+// It returns errors.ErrTransportRetryable to indicate if the request is consider as retryable.
+func (e *ert) RoundTrip(req *http.Request) (*http.Response, error) {
 	if e.bo == nil {
 		return e.roundTrip(req)
 	}
-	var r interface{}
-	r, err = e.bo.Do(req.Context(), func() (interface{}, error) {
+
+	r, err := e.bo.Do(req.Context(), func() (interface{}, error) {
```
kpango, line 54: why stop using named return variable? which improves performance, why?

internal/net/http/transport/roundtrip.go
```diff
@@ -41,50 +42,66 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
 		http.StatusMovedPermanently,
 		http.StatusBadGateway,
 		http.StatusGatewayTimeout:
-		return nil, errors.ErrTransportRetryable
+		return true
+	}
+	return false
+}
+
+func closeBody(res *http.Response) {
```
kpango, line 100:
```suggestion
func closeBody(body *io.ReadCloser) {
```

internal/net/http/transport/roundtrip.go
```diff
@@ -41,50 +42,66 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
-	res, err = e.transport.RoundTrip(req)
-	if err == nil {
-		return res, nil
-	}
-	if res == nil {
-		return nil, err
-	}
-	_, err = io.Copy(ioutil.Discard, res.Body)
+func (e *ert) roundTrip(req *http.Request) (*http.Response, error) {
+	res, err := e.transport.RoundTrip(req)
```
kpango, line 69: same

internal/net/http/transport/roundtrip.go
kevindiu: I am not sure too. Seems the retry mechanism is not implemented yet? 
I am thinking that we cannot stop the retry mechanism in backoff unless we cancel the context.
@kpango what do you think?

kpango: ErrTransportRetryable is new feature of this PR I think we should strictly check error is Retryable or not and if err is not ErrTransportRetryable we should pass error outside of backoff

kevindiu: @kpango do you mean we need to update backoff implementation to check the retryable status?

kpango: no I mean you should check error is retryable or not in the Do function

kpango: ErrTransportRetryable is only for transport

kevindiu: how can we control the backoff retry logic here?
https://github.com/vdaas/vald/blob/master/internal/backoff/backoff.go#L80

kpango: no you shouldn't change backoff.go you can change this line
```suggestion
	res, err := e.roundTrip(req)
	if err != nil {
		if err == errors.ErrTransportRetryable {
			return nil,  errors.ErrTransportRetryable
		}
		// write err detailt to outside err variable eg. boerr
		return nil, nil // e.bo.Do returns nil, nil but we have to check boerr
	}
	return res, nil
```

kevindiu: Thinking we can add a new function to backoff to handle the retry mechanism.

kevindiu: @kpango @rinx  I have pushed the changes here. Thank you :)
I will implement test case if this changes is good.
https://github.com/vdaas/vald/pull/589/commits/7a186a806d4400b2092b7add9f7fbdd5710c1354

kevindiu: We can add a new error called `ErrBackoffRetryable` and check the error in backoff implementation. 
If you support my suggestion I will create a new proposal on SET project board first :)

kpango: thank you for your suggestion, please make design proposal for backoff package.
I agree with your opinion, I think it's better to handle by boolean for example
change Do interface to 
```
Do(ctx context.Context, fn func() (val interface{}, retryable bool, err error))
```

kevindiu: thank you for your comment. I think this is better maybe 🤔
`Do(ctx context.Context, fn func() (val interface{}, status backoff.Status, err error))`

kpango: I think it's boring to define specific type for backoff enable/disable because there is only 2 choice
if we define like my idea users can use it easly.
```
r, err := e.bo.Do(req.Context(), func() (interface{}, bool, error) {
	res, err :=  e.roundTrip(req)
        return res, err == errors. ErrTransportRetryable, err
```
or 
```
r, err := e.bo.Do(req.Context(), func() (interface{}, bool, error) {
	res, err :=  e.roundTrip(req)
        return res, err != nil, err
```

kpango: by the way, please send design proposals we should discussed there

kevindiu: you are right, I will create a proposal first and lets discuss the design there.
In this PR I will not include that new feature to the backoff first, and I will implement new test code base on the current implement.

internal/net/http/transport/roundtrip.go
```diff
@@ -51,12 +51,26 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
...
+			// if the error is not retryable, return nil error to backoff to terminate the backoff execution
+			boErr = err
+			return nil, nil
+		}
+		return res, nil
 	})
 	if err != nil {
 		return nil, err
 	}
+	if boErr != nil {
```
rinx, line 71: Thanks! it's better than before.
however the naming of this variable is a little bit strange. `boErr` is not an error of backoff.

internal/net/http/transport/roundtrip.go
```diff
@@ -41,20 +42,37 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
+		return nil, boErr
```
kevindiu, line 72:
```suggestion
	var ferr error
	r, err := e.bo.Do(req.Context(), func() (res interface{}, err error) {
		res, err = e.roundTrip(req)
		if err != nil {
			// if the error is retryable, return the error and let backoff to retry.
			if errors.Is(err, errors.ErrTransportRetryable) {
				return nil, err
			}
			// if the error is not retryable, return nil error to backoff to terminate the backoff execution
			ferr = err
			return nil, nil
		}
		return res, nil
	})
	if err != nil {
		return nil, err
	}
	if ferr != nil {
		return nil, ferr
```

internal/net/http/transport/roundtrip.go
kevindiu: what do you think about this?
https://github.com/vdaas/vald/projects/3#card-42651099

rinx: no.1 seems to be the better choice.
There are already several lines of codes that checks the returned error is retryable or not in agent-sidecar codes. so please refactor them at the same time.

internal/net/http/transport/roundtrip.go
vankichi: @kpango @rinx FYI
https://github.com/vdaas/vald/projects/3#card-42651099

internal/net/http/transport/roundtrip.go
```diff
@@ -41,20 +42,37 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
 func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 	if e.bo == nil {
 		return e.roundTrip(req)
 	}
-	var r interface{}
-	r, err = e.bo.Do(req.Context(), func() (interface{}, error) {
-		return e.roundTrip(req)
+
+	var ferr error
+	r, err := e.bo.Do(req.Context(), func() (res interface{}, err error) {
```
hlts2, line 55: could you please fix to named return.

```
var r interface{}
r, err = e.bo.Do
```

internal/net/http/transport/roundtrip.go
```diff
@@ -41,20 +42,37 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
 	if err != nil {
 		return nil, err
 	}
+	if ferr != nil {
+		return nil, ferr
+	}
 
-	var ok bool
-	res, ok = r.(*http.Response)
+	res, ok := r.(*http.Response)
```
hlts2, line 75: could you please use named return value?

internal/net/http/transport/roundtrip.go
kevindiu: do you mean this?
```suggestion
	res, err = e.bo.Do(req.Context(), func() (res interface{}, err error) {
```

internal/net/http/transport/roundtrip.go
kevindiu: I think it use the named return value. it think it will just create a new instance of `ok bool` but do no create `res`

internal/net/http/transport/roundtrip.go
hlts2: yes!

hlts2: no.
```
func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
	if e.bo == nil {
		return e.roundTrip(req)
	}

	var ferr error
	var r interface{}
	r, err = e.bo.Do(req.Context(), func() (res interface{}, err error) {
		....
	})
	if err != nil {
		return nil, err
	}
	if ferr != nil {
		return nil, ferr
	}
 
	var ok bool
	res, ok = r.(*http.Response)
	if !ok {
		return nil, errors.ErrInvalidTypeConversion(r, res)
	}
	return res, nil
}
```

kevindiu: I think we cannot do it as `bo.Do` returns `interface{}`

internal/net/http/transport/roundtrip.go
hlts2: https://github.com/vdaas/vald/pull/589/#discussion_r462709330

kevindiu: so it will create a new variable?

kevindiu: https://play.golang.org/p/M9-KGxWKdt-

internal/net/http/transport/roundtrip.go
kevindiu: commented below

kpango: hmm, also I think we can reduce reflection 
stop using named return in Do function & directry insert to outer named return res variable

kevindiu: I also thought so. Personally I don't like named return value as I think it is confusing XDD

kpango: but it reduce memory allocation time and It's memory efficient.
I think we should use named return by default, but we should be careful about naming for each return values

kevindiu: yes :( we need to be careful when we use it, especially in multi-layer functions

kevindiu: if I change the implementation below, the return response will becomes nil, do you have any idea?
```golang
 49 func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 50     if e.bo == nil {
 51         return e.roundTrip(req)
 52     }
 53
 54     var ferr error
 55     _, err = e.bo.Do(req.Context(), func() (interface{}, error) {
 56         res, err = e.roundTrip(req)
 57         if err != nil {
 58             // if the error is retryable, return the error and let backoff to retry.
 59             if errors.Is(err, errors.ErrTransportRetryable) {
 60                 return nil, err
 61             }
 62             // if the error is not retryable, return nil error to terminate the backoff execution
 63             ferr = err
 64             return nil, nil
 65         }
 66         return res, nil
 67     })
 68     if err != nil {
 69         return nil, err
 70     }
 71     if ferr != nil {
 72         return nil, ferr
 73     }
 74
 75     /*
 76         res, ok := r.(*http.Response)
 77         if !ok {
 78             return nil, errors.ErrInvalidTypeConversion(r, res)
 79         }
 80     */
 81
 82     return res, nil
 83 }
```

kevindiu: sorry it made some mistake in my test code, I have fixed it.

internal/net/http/transport/roundtrip.go
```diff
@@ -51,31 +51,28 @@ func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 		return e.roundTrip(req)
 	}
 
-	var ferr error
-	r, err := e.bo.Do(req.Context(), func() (res interface{}, err error) {
-		res, err = e.roundTrip(req)
-		if err != nil {
+	var boerr error
```
kpango, line 54: I think it’s not boerrr, please name it as rterr

internal/net/http/transport/roundtrip.go
```diff
@@ -41,50 +42,77 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
+// It returns errors.ErrTransportRetryable to indicate if the request is consider as retryable.
 func (e *ert) RoundTrip(req *http.Request) (res *http.Response, err error) {
 	if e.bo == nil {
 		return e.roundTrip(req)
 	}
-	var r interface{}
-	r, err = e.bo.Do(req.Context(), func() (interface{}, error) {
-		return e.roundTrip(req)
+
+	var fnerr error
```
hlts2, line 54: I think `rerr` is better too.
https://github.com/vdaas/vald/pull/589/files#discussion_r463391391

kevindiu: thank you. I think i misunderstood `rt` as `return` 😂 it means `round trip`

internal/net/http/transport/roundtrip.go
```diff
@@ -41,50 +42,77 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
 	if e.bo == nil {
 		return e.roundTrip(req)
 	}
-	var r interface{}
-	r, err = e.bo.Do(req.Context(), func() (interface{}, error) {
-		return e.roundTrip(req)
+
+	var boerr error
+	_, err = e.bo.Do(req.Context(), func() (interface{}, error) {
+		r, e := e.roundTrip(req)
```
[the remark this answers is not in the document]
kpango: please do not overwrite e we already use e for receiver variable

internal/net/http/transport/roundtrip.go
```diff
@@ -41,50 +42,77 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
-	if res == nil {
-		return nil, err
-	}
-	_, err = io.Copy(ioutil.Discard, res.Body)
 	if err != nil {
-		log.Error(err)
+		if res != nil { // just in case we check the response as it depends on RoundTrip impl.
+			closeBody(res.Body)
+			if retryable(res) {
+				return nil, errors.Wrap(errors.ErrTransportRetryable, err.Error())
```
vankichi, line 85: Does it need to output error log before return?

kevindiu: or maybe it is better to write it on line 69 and 72?

internal/net/http/transport/roundtrip.go
```diff
@@ -41,50 +42,77 @@ func NewExpBackoff(opts ...Option) http.RoundTripper {
...
-		log.Error(err)
+
+	if res != nil && retryable(res) {
+		closeBody(res.Body)
+		return nil, errors.ErrTransportRetryable
 	}
+	return res, nil
+}
+
+func retryable(res *http.Response) bool {
```
kpango, line 98: I think it's better to pass status code only instead of http.Response pointer

kevindiu: I think it is different way of thinking. `is the HTTP response retryable` or `is that status code retryable`.
I think the first one is better, what do you think?

internal/net/http/transport/roundtrip.go
vankichi: umm, in my opinion, I prefer output between L84 and L85.

internal/net/http/transport/roundtrip.go
kevindiu: fixed

internal/net/http/transport/roundtrip.go
kevindiu: or line 81?

kevindiu: line 84 may not log the error if the response is not retryable. 🤔

vankichi: so between L81 and L82 is better?

kevindiu: I think yes.

vankichi: ok

kevindiu: fixed

internal/net/http/transport/roundtrip_test.go
```diff
@@ -263,44 +135,165 @@ func Test_ert_RoundTrip(t *testing.T) {
...
+					},
+				},
+			},
+			want: want{
+				err: errors.Wrap(errors.ErrTransportRetryable, "error"),
+			},
+		},
 	}
 
+	log.Init()
```
hlts2, line 293: It panics depending on the order of the test cases that call log.Init. So could you please add it to TestMain.?

kevindiu: Oh I didn't know that before. fixed.

PR #612, opened 2020-08-05
implement billion scale data

kmrmt:
### Description:

implement billion scale dataset loader and download job.
If you have a good package name, please suggest that.

### Related Issue:

Nothing.

There is not test but benchmark.

### Environment:

- Go Version: 1.14.4
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.0

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [X] Update benchmarks and performances [type/bench]
- [x] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


Makefile.d/bench.mk
```diff
@@ -45,6 +63,28 @@ bench/datasets/md5dir/print:
...
+	$(SIFT1B_QUERY_FILE) \
+	$(SIFT1B_GROUNDTRUTH_DIR)
+
+.PHONY: bench/dataset/large/deep1b
+## fetch deep1b dataset for benchmark
+bench/dataset/large/deep1b: \
+	$(DEEP1B_BASE_FILE) \
+	$(DEEP1B_LEARN_FILE) \
+	$(DEEP1B_QUERY_FILE) \
+	$(DEEP1B_GROUNDTRUTH_FILE)
```
rinx, line 86: How about to rename them like this?
```suggestion
.PHONY: bench/datasets/large
## fetch large datasets for benchmark
bench/datasets/large: \
	bench/dataset/large/sift1b \
	bench/dataset/large/deep1b

.PHONY: bench/datasets/large/sift1b
## fetch sift1b dataset for benchmark
bench/datasets/large/sift1b: \
	$(SIFT1B_BASE_FILE) \
	$(SIFT1B_LEARN_FILE) \
	$(SIFT1B_QUERY_FILE) \
	$(SIFT1B_GROUNDTRUTH_DIR)

.PHONY: bench/datasets/large/deep1b
## fetch deep1b dataset for benchmark
bench/datasets/large/deep1b: \
	$(DEEP1B_BASE_FILE) \
	$(DEEP1B_LEARN_FILE) \
	$(DEEP1B_QUERY_FILE) \
	$(DEEP1B_GROUNDTRUTH_FILE)
```

kmrmt: You are right. I fixed it.

kmrmt:
@hlts2 I want your advice how way to fix e2e benchmark.
I fixed Dataset interface to have same one small dataset and large dataset.
In this case, it is difficult for me to fix the bulk_insert benchmark.

hack/benchmark/assets/x1b/loader.go
```diff
@@ -0,0 +1,195 @@
...
+	*file
+}
+
+func open(fname string, elementSize int) (f *file, err error) {
+	fp, err := os.Open(fname)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		err = fp.Close()
```
hlts2, line 81: When an error occurs in the `syscall.Mmap` and `fp.Close` returns nil,  I think error will disappear.
So I think we should use `errors.Wrap`

hack/benchmark/assets/x1b/loader_test_bench.go
```diff
@@ -0,0 +1,135 @@
...
+func BenchmarkFVecs(b *testing.B) {
+	fv, err := NewFVecs(fvecsFile)
+	defer func() {
+		if err := fv.Close(); err != nil {
+			b.Fatal(err)
+		}
+	}()
+	if err != nil {
+		b.Fatal(err)
+	}
```
hlts2, line 73:
```suggestion
	fv, err := NewFVecs(fvecsFile)
	if err != nil {
		b.Fatal(err)
	}
	defer func() {
		if err := fv.Close(); err != nil {
			b.Fatal(err)
		}
	}()
```

hack/benchmark/core/benchmark/strategy/bulk_insert.go
```diff
@@ -26,18 +26,56 @@ import (
...
+				v := make([][]float32, 0, size)
+				for i := 0; i < size; i++ {
+					arr, err := dataset.Train(i)
+					if err != nil {
+						break
+					}
+					v = append(v, arr.([]float32))
+				}
+				b.StartTimer()
+				defer b.StopTimer()
```
hlts2, line 54: I am sorry. if I make misstake.
```suggestion
		bb.StopTimer()
		bb.ReportAllocs()
		bb.ResetTimer()
		bb.StartTimer()

```

hlts2 (changes requested):

hack/benchmark/core/benchmark/strategy/bulk_insert.go
kmrmt: The argument is **b *testing.B**, so your suggestion is typo, isn't it?

hlts2: sorry. My suggestion is typo.
b.〇〇Timer is correct:bow:

kmrmt: Thank you! I fix it soon.

hack/benchmark/assets/x1b/loader_test_bench.go
```diff
@@ -0,0 +1,134 @@
...
+				if err := bv.Close(); err != nil {
+					bb.Fatal(err)
+				}
+				bv, err = NewBVecs(bvecsFile)
+				i = 0
+			}
+			if err != nil {
+				bb.Fatal(err)
+			}
+			bb.Log(v)
```
hlts2, line 59: Do we need this output?
I think the output of vector can fill the screen:thinking:

hlts2 (changes requested):

hack/benchmark/assets/x1b/loader.go
```diff
@@ -0,0 +1,195 @@
...
+	Size() int
+	Close() error
+}
+
+type Bvecs interface {
+	X1b
+	LoadUint8(i int) ([]uint8, error)
+}
+
+type Fvecs interface {
```
kevindiu, line 48: I think the naming of this interface is not good.
I suggest rename this to `FloatVector`

hack/benchmark/assets/x1b/loader.go
```diff
@@ -0,0 +1,195 @@
...
+const (
+	headerSize = 4
+)
+
+var (
+	ErrOutOfBounds         = errors.New("out of bounds")
+	ErrUnsupportedFileType = errors.New("unsupported file type")
+)
+
+type X1b interface {
```
kevindiu, line 36: what is this interface means?

kmrmt: It means operating billion scale dataset interface. These dataset names are sift1b and deep1b, so they have common sub strings "1b".
However, I think that it is not the best name.

hack/benchmark/assets/x1b/loader_test_bench.go
```diff
@@ -0,0 +1,134 @@
...
+			v, err := bv.Load(i)
+			switch err {
+			case nil:
+				i++
+				continue
+			case ErrOutOfBounds:
+				if err := bv.Close(); err != nil {
+					bb.Fatal(err)
+				}
+				bv, err = NewBVecs(bvecsFile)
```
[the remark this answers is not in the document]
kmrmt: Sorry, I forgot changing the implementations, so I can implement simpler.
I will fix it.

hack/benchmark/core/benchmark/strategy/bulk_insert.go
```diff
@@ -26,18 +26,62 @@ import (
...
+						return maxBulkSize
+					} else {
+						return dataset.TrainSize()
+					}
+				}()
+				v := make([][]float32, 0, size)
+				for i := 0; i < size; i++ {
+					arr, err := dataset.Train(i)
+					if err != nil {
+						break
```
kevindiu, line 49: will we miss some data if we break the loop here?

hack/benchmark/core/benchmark/strategy/util.go
```diff
@@ -55,18 +65,28 @@ func insertAndCreateIndex32(ctx context.Context, c core.Core32, dataset assets.D
...
+	ids = make([]uint, 0, dataset.TrainSize()*bulkInsertCnt)
 
+	n := 0
 	for i := 0; i < bulkInsertCnt; i++ {
+		train := make([][]float64, 0, dataset.TrainSize()/bulkInsertCnt)
+		for j := 0; j < len(train); j++ {
+			v, err := dataset.Train(n)
+			if err != nil {
+				n = 0
+				break
```
kevindiu, line 85: same as above.
`will we miss some data if we break the loop here?`

hack/benchmark/internal/e2e/strategy/insert.go
```diff
@@ -50,14 +51,17 @@ func (isrt *insert) Run(ctx context.Context, b *testing.B, c client.Client, data
...
 		bb.StopTimer()
 		bb.ReportAllocs()
 		bb.ResetTimer()
 		bb.StartTimer()
 		for i := 0; i < bb.N; i++ {
-			isrt.do(ctx, bb, c, ids[cnt%len(ids)], train[cnt%len(train)])
+			v, err := dataset.Train(cnt % dataset.TrainSize())
+			if err != nil {
+				cnt = 0
+				break
```
kevindiu, line 62: same as above.
`will we miss some data if we break the loop here?`

hack/benchmark/core/benchmark/strategy/bulk_insert.go
kmrmt: In almost cases err is nil because Train function returns only ErrOutOfBounds as error and size is less or equal train dataset size.
I think this sentence is redundant now, but Train may changes returning other error.
So that, it might be better dividing the process that err is ErrOutOfBounds and the others.

hlts2 (dismissed):
LGTM

hack/benchmark/core/benchmark/strategy/bulk_insert.go
kevindiu: I mean the data may not be loaded completely whenever the error is occurred, so I think there is a case that the error occurred and the `arr` is not append to `v` in [line 51](https://github.com/vdaas/vald/pull/612/files#diff-408799d8c77478a76c2fa1f9d331d158R51).

What do you think?

kmrmt: Sorry, I didn't think seriously about it. You are right.
The best way I think is changing WithProp32 interface like this and WithProp64 in the same way.
`func(ctx context.Context, b *testing.B, c core.Core32, [][]float32, ids []uint, cnt *uint64) (interface{}, error)`
The good point is what enables to use same [][]float32 data in all benchmarks. We can get fair benchmark in the same benchmark using the same data, if error is occurred. (However, We may get the unfair result in the different benchmark.)
The bad point is what changes interfaces and needs to implement data generation. It could be a complex implementation.

kevindiu: I am not sure if we can do this? to train the dataset from index 0 again whenever error occurred 
```go
				for i := 0; i < size;  {
					arr, err := dataset.Train(i % size)
					if err == nil {
					         v = append(v, arr.([]float32))
                                                  i++
					}
				}
```

kevindiu: what I concerned was that the testing result may inaccurate when error is occurred

kmrmt: Sorry, I sent it halfway through.
If my proposal is chosen, I think that it is good to change it in the new pull request.

kmrmt: Sorry, I have reconsidered.
Benchmark is important to use same data for comparing result.
So, my implementation is bad, and it might be good to stop the benchmark if error is occurred.
How about it?

kevindiu: `it might be good to stop the benchmark if error is occurred`
I think it is ok :) 👍

kmrmt: Thank you for your comment! I will fix to stop the benchmark.

kmrmt:
fix benchmark and rebase

hack/benchmark/assets/x1b/loader_bench_test.go
```diff
@@ -0,0 +1,113 @@
...
+			switch err {
+			case nil:
+				i++
+			case ErrOutOfBounds:
+				i = 0
+			default:
+				bb.Fatal(err)
+			}
+		}
+	})
```
kevindiu, line 54: I think it has the same problem: some data may lost when error occurs.

```suggestion
	b.Run(bvecsFile, func(bb *testing.B) {
		bb.ReportAllocs()
		bb.ResetTimer()
		
		i := 0
		for n := 0; n < bb.N; {
			_, err := bv.Load(i)
			switch err {
			case nil:
				i++
				n++
			case ErrOutOfBounds:
				i = 0
			default:
				bb.Fatal(err)
			}
		}
	})
```

hack/benchmark/assets/x1b/loader_bench_test.go
```diff
@@ -0,0 +1,113 @@
...
+		if err := fv.Close(); err != nil {
+			b.Fatal(err)
+		}
+	}()
+
+	i := 0
+	b.Run(fvecsFile, func(bb *testing.B) {
+		bb.ReportAllocs()
+		bb.ResetTimer()
+		for n := 0; n < bb.N; n++ {
```
kevindiu, line 72: same

hack/benchmark/assets/x1b/loader_bench_test.go
```diff
@@ -0,0 +1,113 @@
...
+		b.Fatal(err)
+	}
+	defer func() {
+		if err := iv.Close(); err != nil {
+			b.Fatal(err)
+		}
+	}()
+
+	i := 0
+	b.Run(ivecsFile, func(bb *testing.B) {
```
kevindiu, line 98: same

hack/benchmark/assets/x1b/loader_bench_test.go
kmrmt: If the error "ErrOutOfBounds" is occurred, the whole data is loaded and it should continue the benchmark from the head.
If the others are occurred, they stops the benchmark.

kevindiu: Sorry if i am wrong here.
The problem is that the `n++` will still be executed even when the error occurred, so the number of times `bv.Load` executed maybe less than `bb.N`.

kmrmt: I see. I got a good idea for this code.
In this code ErrOutOfBounds is never occurred and I think it seems easy to understand the codes.
However, % operation takes much time.
```
		i := 0
		size := bv.Size()
		for n := 0; n < bb.N; n++ {
			_, err := bv.Load(i % size)
			if err != nil {
				bb.Fatal(err)
			}
			i++
		}
```

kevindiu: I think it is good!

kevindiu (approved):
LGTM

hlts2 (dismissed):
LGTM

vankichi (dismissed):

PR #614, opened 2020-08-06
Coding guideline: Add error checking section

kevindiu:
### Description:

This PR add a error checking section to the coding guideline.

### Environment:

- Go Version: 1.14.4
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.0

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [x] Add documents [type/documentation]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/docs/contributing/coding-style.md
+++ b/docs/contributing/coding-style.md
@@ -299,7 +299,7 @@ Please use [internal/errgroup](https://github.com/vdaas/vald/blob/master/interna
 
 ### Error checking
 
-All functions returns `error` if the function have possiblility to fail. It is very important to ensure the error checking is performed.
+All functions return `error` if the function can fail. It is very important to ensure the error checking is performed.
 To reduce human mistake that missing the error checking, please check the error using the following style:
 
 ```golang
--- a/docs/contributing/coding-style.md
+++ b/docs/contributing/coding-style.md
@@ -221,7 +221,7 @@ ErrInvalidCacherType = New("invalid cacher type")
 
 ### Methods
 
-The method name should be named as:
+In this section, rules also apply to the `function` (without receiver). The method name should be named as:
 
 - Use MixedCaps.
 
@@ -236,6 +236,7 @@ func (s *something) some_method() {}
 func (s *something) someMethod() {}
 ```
 
+- Avoid using long function name.
 - Do not use short form unless the function name is too long.
 
 ```go
@@ -272,7 +273,7 @@ So please delete the unused variable.
 Generally, the unused variable should be reported during compilation, but in some cases, the compiler may not report an error.
 This is an example of the unused variable declaration that does not cause a compilation error.
 
-```golang
+```go
 // In this case, this example are not using `port` field, but dose not cause a compilation error.
 // So please delete `port` field of `server`.
 
@@ -302,7 +303,7 @@ Please use [internal/errgroup](https://github.com/vdaas/vald/blob/master/interna
 All functions return `error` if the function can fail. It is very important to ensure the error checking is performed.
 To reduce human mistake that missing the error checking, please check the error using the following style:
 
-```golang
+```go
 // good
 if err := fn(); err != nil {
     // handle error
@@ -311,7 +312,7 @@ if err := fn(); err != nil {
 
 Instead of this style.
 
-```golang
+```go
 // bad
 err := fn()
 if err != nil {
@@ -321,7 +322,7 @@ if err != nil {
 
 If you need the value outside the if statement, please use the following style:
 
-```golang
+```go
 // good
 conn, err := net.Dial("tcp", "localhost:80")
 if err != nil {
@@ -333,7 +334,7 @@ if err != nil {
 
 Instead of this style.
 
-```golang
+```go
 // bad
 if conn, err := net.Dial("tcp", "localhost:80");  err != nil {
     // handle error
@@ -612,7 +613,7 @@ We do not suggest to modify the generated code other than the `tests` variable, 
     For example, Vald uses [glg](https://github.com/kpango/glg) library for logging by default, if the logger is not initialized before the test, the nil pointer error may be thrown during the test is running.
     You may need to implement `init()` function like:
 
-    ```golang
+    ```go
     func init() {
         log.Init()
     }
@@ -626,7 +627,7 @@ We do not suggest to modify the generated code other than the `tests` variable, 
     Sometimes you may want to skip the detection, for example, Vald uses [fastime](https://github.com/kpango/fastime) library but the internal Goroutine is not closed due to the needs of the library. 
     To skip the goleak detection we need to create the following variable to store the ignore function.
 
-    ```golang
+    ```go
     var (
         // Goroutine leak is detected by `fastime`, but it should be ignored in the test because it is an external package.
         goleakIgnoreOptions = []goleak.Option{
@@ -637,7 +638,7 @@ We do not suggest to modify the generated code other than the `tests` variable, 
 
     And modify the generated test code.
 
-    ```golang
+    ```go
     // before
     for _, test := range tests {
         t.Run(test.name, func(tt *testing.T) {
@@ -657,7 +658,7 @@ We do not suggest to modify the generated code other than the `tests` variable, 
 
     For example:
 
-    ```golang
+    ```go
     for _, test := range tests {
       t.Run(test.name, func(tt *testing.T) {
         defer goleak.VerifyNone(tt, goleakIgnoreOptions...)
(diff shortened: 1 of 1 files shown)


docs/contributing/coding-style.md
```diff
@@ -297,6 +297,51 @@ All errors should define in [internal/errors package](https://github.com/vdaas/v
 
 Please use [internal/errgroup](https://github.com/vdaas/vald/blob/master/internal/errgroup) for synchronized error handling on multi-goroutine processing.
 
+### Error checking
+
+All functions return `error` if the function can fail. It is very important to ensure the error checking is performed.
+To reduce human mistake that missing the error checking, please check the error using the following style:
+
+```golang
```
hlts2, line 305:
```suggestion
```go
```

docs/contributing/coding-style.md
```diff
@@ -297,6 +297,51 @@ All errors should define in [internal/errors package](https://github.com/vdaas/v
...
+// good
+if err := fn(); err != nil {
+    // handle error
+}
+```
+
+Instead of this style.
+
+```golang
```
hlts2, line 314:
```suggestion
```go
```

docs/contributing/coding-style.md
```diff
@@ -297,6 +297,51 @@ All errors should define in [internal/errors package](https://github.com/vdaas/v
...
+// bad
+err := fn()
```
hlts2, line 316: I can understand your opinion.
but It is also good to write that the following is correct when the function name is long.

```go
// good
err := thisHasAReallyLongSuperDescriptiveNameForSomeCrazyReason()
if err != nil {

}

```

docs/contributing/coding-style.md
```diff
@@ -297,6 +297,51 @@ All errors should define in [internal/errors package](https://github.com/vdaas/v
...
+if err != nil {
+    // handle error
+}
+```
+
+If you need the value outside the if statement, please use the following style:
+
+```golang
```
hlts2, line 324:
```suggestion
```go
```

docs/contributing/coding-style.md
```diff
@@ -297,6 +297,51 @@ All errors should define in [internal/errors package](https://github.com/vdaas/v
...
+if err != nil {
+    // handle error
+}
+
+// use the conn
+```
+
+Instead of this style.
+
+```golang
```
hlts2, line 336:
```suggestion
```go
```

docs/contributing/coding-style.md
```diff
@@ -297,6 +297,51 @@ All errors should define in [internal/errors package](https://github.com/vdaas/v
...
 Please use [internal/errgroup](https://github.com/vdaas/vald/blob/master/internal/errgroup) for synchronized error handling on multi-goroutine processing.
 
+### Error checking
+
+All functions return `error` if the function can fail. It is very important to ensure the error checking is performed.
+To reduce human mistake that missing the error checking, please check the error using the following style:
+
+```golang
+// good
+if err := fn(); err != nil {
```
hlts2, line 307: this code is correct when there is only one error handling.
but when there is two error handling, I think the following is correct.

```go
err := fn()
if err != nil {

}

err = fn2()
if err != nil {

}

```
If you do not reuse the error, the variable will be declared again.

docs/contributing/coding-style.md
```diff
@@ -297,6 +297,51 @@ All errors should define in [internal/errors package](https://github.com/vdaas/v
...
+}
+
+// use the conn
+```
+
+Instead of this style.
+
+```golang
+// bad
+if conn, err := net.Dial("tcp", "localhost:80");  err != nil {
```
hlts2, line 338: I think when logic in the `else` is log, this guideline is correct.
but when logic in the `else` is short, I think the following example is correct.

```
if conn, err := net.Dial(); err != nil {

} else {
    return conn.Close()
}
```

docs/contributing/coding-style.md
kevindiu: how about this? but I think it is ok to create another `err` variable in `fn2()`, what do you think?

```suggestion
err := fn()
if err != nil {
    // handle err
}

if err = fn2(); err != nil {
    // handle err
}
```

docs/contributing/coding-style.md
kevindiu: instead of this strategy, I think we should suggest NOT to declare long function name. What do you think?

docs/contributing/coding-style.md
kevindiu: in your case I think maybe this is better

if conn, err := net.Dial(); err == nil {
    return conn.Close()
}

// next

docs/contributing/coding-style.md
hlts2: ok, it seems good to me.

kevindiu: @vankichi what do you think?
If you support my idea, should I also update the `Methods` section.

kevindiu:
I found that the section name `Method` is wrong, it should be `Function`.
https://github.com/vdaas/vald/blob/af55ed968ae030987586e67775ee30619cd9eff5/docs/contributing/coding-style.md#methods
@hlts2 @vankichi what do you think?

docs/contributing/coding-style.md
vankichi: I think too long term for any methods make us confused in some cases.
we should use short but suitable terms for method, and we should cover using comment if it may occur some misreading.

docs/contributing/coding-style.md
vankichi: @kevindiu 
in my opinion, this rule is not necessary.
And, we should not design many coding rules for each section.
It too hard to remember all of the rules for me 😓

docs/contributing/coding-style.md
vankichi: @kevindiu 
your suggestion seems like having no error handling 🤔

kevindiu: @vankichi error handling is the `next` block.

```go
if conn, err := net.Dial(); err == nil {
    return conn.Close()
}

// error handling
```

docs/contributing/coding-style.md
kevindiu: @vankichi I will update the `Method` section later

docs/contributing/coding-style.md
kevindiu: @vankichi so do you think the original guideline is ok? 
```go
// good
if err := fn(); err != nil {
    // handle error
}
```

docs/contributing/coding-style.md
hlts2: ok
it seems good

vankichi: I'm not sure why we should use this.
IMO, I just use only @hlts2 's suggestion.

docs/contributing/coding-style.md
vankichi: it's ok

vankichi:
Method and Function may be correct?

docs/contributing/coding-style.md
hlts2: > this rule is not necessary.

I think so too.
The more rules we have, the harder it is to learn and the more complicated it becomes.

hlts2:
@kevindiu 

I think this section is correct.
Because Method is a function associated with a "type"  and Functions are general functions (starting with func and no associated with a "type")

cc: @vankichi

docs/contributing/coding-style.md
kevindiu: if the `if block` become huge, it makes code difficult to read. In @hlts2 suggestion,
```go
if conn, err := net.Dial(); err != nil {
    // many
    // lines
    // of
    // logic
} else {
    return conn.Close()
}
```

IMO, it is harder to read than my suggestion.

kevindiu:
@hlts2 thank you for your comment. 
than should we rename the Method section to Method & Function?

docs/contributing/coding-style.md
vankichi: umm...
I could not understand the benefit.
If it's hard to read, we should write as below.
```go
conn, err := net.Dial()
if err != nil {
  // error handling
}
// next
```

vankichi: ah, is it a bad pattern?

kevindiu: I think it is a good pattern, like what I wrote in the document. 
```
322 If you need the value outside the if statement, please use the following style:
323
324 ```go
325 // good
326 conn, err := net.Dial("tcp", "localhost:80")
327 if err != nil {
328     // handle error
329 }
330
331 // use the conn
332 ```
```

Seems the origin question from @hlts2 is not a good suggestions, what do you think?

vankichi: i misunderstood, sorry.

PR #615, opened 2020-08-07
Add internal/net test

kevindiu:
Author review required.
### Description:

This PR implements test case for internal/net/net.go. 
This PR also include the following changes:

Bug fix:
- In ScanPorts(), when `start > end`, unexpected ports will be scanned
   Fix: https://github.com/vdaas/vald/pull/615/files#diff-111853981bfa0f81d6c8945300daf97bR122-R124

- In ScanPorts(), when `end` is 65535, the `ScanPorts()` will never return due to golang bug or behavior
   Fix: https://github.com/vdaas/vald/pull/615/files#diff-111853981bfa0f81d6c8945300daf97bR134
   Ref: https://github.com/vdaas/vald/pull/615#discussion_r469117449

- In ScanPorts(), scanning ports in range, with some ports are failed but some success, the error will be returned instead of returning succeed ports 
   Fix: https://github.com/vdaas/vald/pull/615/files#diff-111853981bfa0f81d6c8945300daf97bR143-R144
          https://github.com/vdaas/vald/pull/615/files#diff-111853981bfa0f81d6c8945300daf97bR163-R165

Refactor:
- In ScanPorts(), do not return connection close error.
   Ref: https://github.com/vdaas/vald/pull/615/files#diff-111853981bfa0f81d6c8945300daf97bR151-R153

- In SplitHostPort(), commented out the `::` support due to invalid implementation, which always return error.
   To support loopback IPv6 address please read the function comment.
   Ref: https://github.com/vdaas/vald/pull/615/files#diff-111853981bfa0f81d6c8945300daf97bR98-R101

### Environment:

- Go Version: 1.14.4
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.0

### Types of changes:

- [x] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/net/net.go
+++ b/internal/net/net.go
@@ -142,6 +142,7 @@ func ScanPorts(ctx context.Context, start, end uint16, host string) (ports []uin
 	eg.Limitation(int(rl.Max) / 2)
 
 	var mu sync.Mutex
+
 	for i := start; i >= start && i <= end; i++ {
 		port := i
 		eg.Go(func() error {
--- a/internal/net/net.go
+++ b/internal/net/net.go
@@ -27,6 +27,8 @@ import (
 	"syscall"
 
 	"github.com/vdaas/vald/internal/errgroup"
+	"github.com/vdaas/vald/internal/errors"
+	"github.com/vdaas/vald/internal/log"
 )
 
 const (
@@ -36,30 +38,38 @@ const (
 	defaultPort = 80
 )
 
-type Conn = net.Conn
-type Dialer = net.Dialer
-type ListenConfig = net.ListenConfig
-type Listener = net.Listener
-type Resolver = net.Resolver
+type (
+	Conn         = net.Conn
+	Dialer       = net.Dialer
+	ListenConfig = net.ListenConfig
+	Listener     = net.Listener
+	Resolver     = net.Resolver
+)
 
 var (
 	DefaultResolver = net.DefaultResolver
 )
 
+// Listen is a wrapper function of the net.Listen function.
 func Listen(network, address string) (Listener, error) {
 	return net.Listen(network, address)
 }
 
+// IsLocal returns if the host is the localhost address.
 func IsLocal(host string) bool {
 	return host == localHost ||
 		host == localIPv4 ||
 		host == localIPv6
 }
 
+// Dial is a wrapper function of the net.Dial function.
 func Dial(network string, addr string) (conn Conn, err error) {
 	return net.Dial(network, addr)
 }
 
+// Parse parses the hostname, IPv4 or IPv6 address and return the hostname/IP, port number,
+// whether the address is IP, and any parsing error occurred.
+// The address should contains the port number, otherwise an error will return.
 func Parse(addr string) (host string, port uint16, isIP bool, err error) {
 	host, port, err = SplitHostPort(addr)
 	isIP = IsIPv6(host) || IsIPv4(host)
@@ -69,18 +79,22 @@ func Parse(addr string) (host string, port uint16, isIP bool, err error) {
 	return host, port, isIP, err
 }
 
+// IsIPv6 returns weather the address is IPv6 address.
 func IsIPv6(addr string) bool {
 	return net.ParseIP(addr) != nil && strings.Count(addr, ":") >= 2
 }
 
+// IsIPv4 returns weather the address is IPv4 address.
 func IsIPv4(addr string) bool {
 	return net.ParseIP(addr) != nil && strings.Count(addr, ":") < 2
 }
 
+// SplitHostPort splits the address, and return the host/IP address and the port number,
+// and any error occurred.
 func SplitHostPort(hostport string) (host string, port uint16, err error) {
 	switch {
 	case strings.HasPrefix(hostport, "::"):
-		hostport = localIPv6 + hostport
+		hostport = localIPv6 + hostport[2:]
 	case strings.HasPrefix(hostport, ":"):
 		hostport = localIPv4 + hostport
 	}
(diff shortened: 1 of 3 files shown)


internal/net/net.go
```diff
@@ -69,18 +79,22 @@ func Parse(addr string) (host string, port uint16, isIP bool, err error) {
...
 	return net.ParseIP(addr) != nil && strings.Count(addr, ":") < 2
 }
 
+// SplitHostPort splits the address, and return the host/IP address and the port number,
+// and any error occurred.
 func SplitHostPort(hostport string) (host string, port uint16, err error) {
 	switch {
 	case strings.HasPrefix(hostport, "::"):
-		hostport = localIPv6 + hostport
+		hostport = localIPv6 + hostport[2:]
```
kevindiu, line 97: The format of IPv6 address with port number should be `[addr]:port`, the logic of line 96 is incorrect, should I fix it? @kpango @rinx

kevindiu: or if the `hostport` do not contains port number, there are no `[]` to wrap the address, instead the host should use the `hostport` with port `defaultPort`

kevindiu: I suggest:
```go
switch {
case hostport == "::1":
    hostport = "[" + localIPv6 + "]:" + defaultPort
........
```

kevindiu: or we can return directly.
```go
switch {
case hostport == "::1":
    return localIPv6, defaultPort, nil

Comment (kpango) [reply internal/net/net.go:97]: I think we have to make 2 changes, this case was implemented for detecting loopback address, if there is no port definition on hostport string, we can use default port but in this case we have to check there is a port in hostport string or not, and if exists we have to parse it.
one more thing as you mentioned, loopback addr is `::1` but it also can be written as `::` or `0:0:0:0:0:0:0:1`. so we have to check them

Comment (kevindiu) [reply internal/net/net.go:97]: @kpango 
I think this switch statement is checking if we can use the default port for `hostport`.  
For the case where the `hostport` string contains port number, I think we can rely on `net.SplitHostPort()` in line 101.

For the case without port number, I think we need to check if it is a valid IPv6 address, and return the default port number.

So I think it is correct, what do you think?
```go
switch{
case !strings.HasPrefix(hostport, "[") && IsIPv6(hostport):
     return hostport, defaultPort, nil
.......
```

kpango: could you please read RFC more carefuly. In RFC5952 (https://tools.ietf.org/html/rfc5952#section-6) there is 6 types of ipv6 w/ port format we should support them.
you code seems support only 1 format

kevindiu: @kpango 
May I confirm that this switch case statement is checking if `hostport` contains hostname or port number only and use the default value if not exists? 

For example, line 98-99 use the default local IPv4 address if the host is missing.
I think IPv6 do not have `empty host` case so we don't need to support it.

But I think IPv6 have `empty port`, so I wrote the logic above.

BTW `net.SplitHostPort` doesn't support all RFC format.
https://golang.org/src/net/ipsock.go?s=4766:4832#L146

kevindiu: how should we work on this?

vankichi:
```suggestion
		// hostport = localIPv6 + hostport[2:]
```

vankichi: https://github.com/vdaas/vald/projects/3#card-43504189

internal/net/net.go
```diff
@@ -98,16 +116,22 @@ func SplitHostPort(hostport string) (host string, port uint16, err error) {
...
-	if err != nil {
+	if err = syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rl); err != nil {
 		return nil, err
 	}
 	eg, egctx := errgroup.New(ctx)
 	eg.Limitation(int(rl.Max) / 2)
+
 	var mu sync.Mutex
-	for i := start; i <= end; i++ {
+	for i := start; i >= start && i <= end; i++ {
```
vankichi, line 134: Could you show me the reason?

kevindiu: The valid port number is from 1 to 65535.
it is because when `i = 65534` and execute `i++`, which add i to 65535, the uint16 overflow and reset `i` to 0, causing infinity loop. 
This checking checks if the `i` overflow and prevent the infinity loop.

kpango: It's so complicated problem, the variable `start` and `end` is uint16 variable so `i` becomes uint16.
 the case of `i == math.MaxUint16` then `i` will continue to next incremental loop due to address overflow `i` becomes 0 and continue to loop end

kevindiu: you can reproduce it here
https://play.golang.org/p/Z-Nvjo7Ewu0

kevindiu:
Please ignore the deepsource error

internal/net/net.go
```diff
@@ -36,30 +38,38 @@ const (
...
-type Dialer = net.Dialer
-type ListenConfig = net.ListenConfig
-type Listener = net.Listener
-type Resolver = net.Resolver
+type (
+	Conn         = net.Conn
+	Dialer       = net.Dialer
+	ListenConfig = net.ListenConfig
+	Listener     = net.Listener
+	Resolver     = net.Resolver
```
vankichi, line 46: Could you please add comment for each type? there are global.

internal/net/net.go
```diff
@@ -36,30 +38,38 @@ const (
...
+)
 
 var (
 	DefaultResolver = net.DefaultResolver
```
vankichi, line 50: same ⬆️

internal/net/net.go
kevindiu, line 46:
```suggestion
	// Conn is a wrapper of net.Conn
	Conn         = net.Conn
	
	// Dialer is a wrapper of net.Dialer
	Dialer       = net.Dialer
	
	// ListenConfig is a wrapper of net.ListenConfig
	ListenConfig = net.ListenConfig
	
	// Listener is a wrapper of net.Listener
	Listener     = net.Listener
	
	// Resolver is a wrapper of net.Resolver
	Resolver     = net.Resolver
```

internal/net/net.go
kevindiu:
```suggestion
	// DefaultResolver is a wrapper of net.DefaultResolver
	DefaultResolver = net.DefaultResolver
```

internal/net/net.go
kevindiu: thank you for your comment. how about this?
https://github.com/vdaas/vald/pull/615/files#r469718324

kpango: maybe the word `alias` is suitable than `wrapper`

vankichi: seems good

PR #630, opened 2020-08-19
add internal/s3/reader/option test

vankichi:

### Description:

- add test for `internal/db/storage/blob/s3/reader/option`.
- add comment for each exported var/function for `internal/db/storage/blob/s3/reader/option`.
- refactor set option logic at `WithBackoffOpts`

### Environment:

- Go Version: 1.14.4
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.0

### Types of changes:

- [ ] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/db/storage/blob/s3/reader/option_test.go
+++ b/internal/db/storage/blob/s3/reader/option_test.go
@@ -333,7 +333,6 @@ func TestWithMaxChunkSize(t *testing.T) {
 					maxChunkSize: 10,
 				},
 			},
-			checkFunc: defaultCheckFunc,
 		},
 		{
 			name: "set success when size is negative number",
@@ -345,14 +344,12 @@ func TestWithMaxChunkSize(t *testing.T) {
 					maxChunkSize: -10,
 				},
 			},
-			checkFunc: defaultCheckFunc,
 		},
 		{
 			name: "set success when size is zero",
 			want: want{
 				obj: new(T),
 			},
-			checkFunc: defaultCheckFunc,
 		},
 	}
 
@@ -412,14 +409,12 @@ func TestWithBackoff(t *testing.T) {
 					backoffEnabled: true,
 				},
 			},
-			checkFunc: defaultCheckFunc,
 		},
 		{
 			name: "set success when backoff enabled is false",
 			want: want{
 				obj: new(T),
 			},
-			checkFunc: defaultCheckFunc,
 		},
 	}
 
@@ -469,7 +464,10 @@ func TestWithBackoffOpts(t *testing.T) {
 			cmp.AllowUnexported(*got),
 			cmp.AllowUnexported(*w.obj),
 			cmp.Comparer(func(want, got []backoff.Option) bool {
-				return reflect.DeepEqual(want, got) || (got != nil && want != nil)
+				return len(got) == len(want)
+			}),
+			cmp.Comparer(func(want, got backoff.Option) bool {
+				return reflect.ValueOf(got).Pointer() == reflect.ValueOf(want).Pointer()
 			}),
 		}
 		if diff := cmp.Diff(w.obj, got, opts...); diff != "" {
@@ -492,7 +490,6 @@ func TestWithBackoffOpts(t *testing.T) {
 						backoffOpts: opts,
 					},
 				},
-				checkFunc: defaultCheckFunc,
 			}
 		}(),
 		func() test {
@@ -514,16 +511,16 @@ func TestWithBackoffOpts(t *testing.T) {
 				beforeFunc: func(args args, r *T) {
 					r.backoffOpts = args.defaultOpts
 				},
-				checkFunc: defaultCheckFunc,
 			}
 		}(),
-		{
-			name: "set success when opts is nil",
-			want: want{
-				obj: new(T),
-			},
-			checkFunc: defaultCheckFunc,
-		},
+		func() test {
+			return test{
+				name: "set success when opts is nil",
+				want: want{
+					obj: new(T),
+				},
+			}
+		}(),
 	}
 
 	for _, test := range tests {
--- a/internal/db/storage/blob/s3/reader/option_test.go
+++ b/internal/db/storage/blob/s3/reader/option_test.go
@@ -17,205 +17,138 @@
 package reader
 
 import (
+	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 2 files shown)


internal/db/storage/blob/s3/reader/option_test.go
```diff
@@ -17,465 +17,534 @@
...
+			name: "set success when size is positive number",
+			args: args{
+				size: 10,
+			},
+			want: want{
+				obj: &T{
+					maxChunkSize: 10,
+				},
+			},
+			checkFunc: defaultCheckFunc,
```
hlts2, line 336:
```suggestion
```

internal/db/storage/blob/s3/reader/option_test.go
```diff
@@ -17,465 +17,534 @@
...
+			name: "set success when size is negative number",
+			args: args{
+				size: -10,
+			},
+			want: want{
+				obj: &T{
+					maxChunkSize: -10,
+				},
+			},
+			checkFunc: defaultCheckFunc,
```
hlts2, line 348:
```suggestion
```

internal/db/storage/blob/s3/reader/option_test.go
```diff
@@ -17,465 +17,534 @@
...
+		},
+		{
+			name: "set success when size is zero",
+			want: want{
+				obj: new(T),
+			},
+			checkFunc: defaultCheckFunc,
```
hlts2, line 355:
```suggestion
```

internal/db/storage/blob/s3/reader/option_test.go
```diff
@@ -17,465 +17,534 @@
...
+			name: "set success when backoff enabled is true",
+			args: args{
+				enabled: true,
+			},
+			want: want{
+				obj: &T{
+					backoffEnabled: true,
+				},
+			},
+			checkFunc: defaultCheckFunc,
```
hlts2, line 415:
```suggestion
```

internal/db/storage/blob/s3/reader/option_test.go
```diff
@@ -17,465 +17,534 @@
...
+		},
+		{
+			name: "set success when backoff enabled is false",
+			want: want{
+				obj: new(T),
+			},
+			checkFunc: defaultCheckFunc,
```
hlts2, line 422:
```suggestion
```

internal/db/storage/blob/s3/reader/option_test.go
```diff
@@ -17,465 +17,534 @@
...
+				name: "set success when opts is not nil and backoffOpts is nil",
+				args: args{
+					opts: opts,
+				},
+				want: want{
+					obj: &T{
+						backoffOpts: opts,
+					},
+				},
+				checkFunc: defaultCheckFunc,
```
hlts2, line 495:
```suggestion
```

internal/db/storage/blob/s3/reader/option_test.go
```diff
@@ -17,465 +17,534 @@
...
+				},
+				want: want{
+					obj: &T{
+						backoffOpts: append(defaultOpts, opts...),
+					},
+				},
+				beforeFunc: func(args args, r *T) {
+					r.backoffOpts = args.defaultOpts
+				},
+				checkFunc: defaultCheckFunc,
```
hlts2, line 517:
```suggestion
```

internal/db/storage/blob/s3/reader/option_test.go
```diff
@@ -17,465 +17,534 @@
...
+			}
+		}(),
+		{
+			name: "set success when opts is nil",
+			want: want{
+				obj: new(T),
+			},
+			checkFunc: defaultCheckFunc,
```
hlts2, line 525:
```suggestion
```

internal/db/storage/blob/s3/reader/option_test.go
```diff
@@ -17,465 +17,534 @@
...
+		checkFunc  func(want, *T) error
+		beforeFunc func(args, *T)
+		afterFunc  func(args, *T)
+	}
+	defaultCheckFunc := func(w want, got *T) error {
+		opts := []cmp.Option{
+			cmp.AllowUnexported(*got),
+			cmp.AllowUnexported(*w.obj),
+			cmp.Comparer(func(want, got []backoff.Option) bool {
+				return reflect.DeepEqual(want, got) || (got != nil && want != nil)
```
hlts2, line 472: if you use `DeepEquals` to slice, `DeepEquals` always returns false.
so you should check want, got length, and check element pointer of Options. 🤔

hlts2 (changes requested):

internal/db/storage/blob/s3/reader/option_test.go
kevindiu: according to the document, deepequal can check the elements in the slice. 
https://golang.org/pkg/reflect/#DeepEqual
```
Slice values are deeply equal when all of the following are true: they are both nil or both non-nil, they have the same length, and either they point to the same initial entry of the same underlying array (that is, &x[0] == &y[0]) or their corresponding elements (up to length) are deeply equal. Note that a non-nil empty slice and a nil slice (for example, []byte{} and []byte(nil)) are not deeply equal.
```

Writing this should be sufficient.
```suggestion
				return reflect.DeepEqual(want, got)
```

vankichi: That's right.
And, I have another problem.
`reflect.DeepEqual` will not work well when running test about `set_success_when_opts_is_not_nil_and_backoffOpts_is_not_nil`.
How can I resolve it?
```bash
--- FAIL: TestWithBackoffOpts (0.00s)
    --- FAIL: TestWithBackoffOpts/set_success_when_opts_is_not_nil_and_backoffOpts_is_not_nil (0.00s)
        option_test.go:540: error = got = &{<nil> <nil>   <nil> <nil> false [0xac2060] 0}, want &{<nil> <nil>   <nil> <nil> false [0xac2060] 0}
FAIL
exit status 1
FAIL	github.com/vdaas/vald/internal/db/storage/blob/s3/reader	0.147s
```

kevindiu: maybe we need a comparator of `backoff.Option` instead of `[]backoff.Option`
```go
cmp.Comparer(func(want, got backoff.Option) bool
```

kevindiu (approved):
LGTM

hlts2 (approved):
LGTM

PR #643, opened 2020-08-25
add internal/compress/lz4 test

vankichi:

### Description:

- Test
    - create test for `internal/compress/lz4`
    - add comment for each exportable var/function for `internal/compress/lz4`
- Refactor
    - change to use named return at `CompressVector()` because to get an `error` which will be cause at defer function.

### Environment:

- Go Version: 1.14.4
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.0

### Types of changes:

- [ ] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/compress/lz4.go
+++ b/internal/compress/lz4.go
@@ -32,6 +32,7 @@ type lz4Compressor struct {
 	lz4              lz4.LZ4
 }
 
+// NewLZ4 returns Compressor implementation.
 func NewLZ4(opts ...LZ4Option) (Compressor, error) {
 	c := &lz4Compressor{
 		lz4: lz4.New(),
@@ -45,7 +46,9 @@ func NewLZ4(opts ...LZ4Option) (Compressor, error) {
 	return c, nil
 }
 
-func (l *lz4Compressor) CompressVector(vector []float32) ([]byte, error) {
+// CompressVector compresses the data given and returns the compressed data.
+// If CompressVector fails, it will return an error.
+func (l *lz4Compressor) CompressVector(vector []float32) (b []byte, err error) {
 	gob, err := l.gobc.CompressVector(vector)
 	if err != nil {
 		return nil, err
@@ -56,6 +59,7 @@ func (l *lz4Compressor) CompressVector(vector []float32) ([]byte, error) {
 	defer func() {
 		cerr := zw.Close()
 		if cerr != nil {
+			b = nil
 			err = errors.Wrap(err, cerr.Error())
 		}
 	}()
@@ -73,6 +77,8 @@ func (l *lz4Compressor) CompressVector(vector []float32) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// DecompressVector decompresses the compressed data and returns the data.
+// If decompress fails, it will return an error.
 func (l *lz4Compressor) DecompressVector(bs []byte) ([]float32, error) {
 	buf := new(bytes.Buffer)
 	zr := l.lz4.NewReader(bytes.NewReader(bs))
@@ -89,6 +95,7 @@ func (l *lz4Compressor) DecompressVector(bs []byte) ([]float32, error) {
 	return vec, nil
 }
 
+// Reader returns io.ReadCloser implementation.
 func (l *lz4Compressor) Reader(src io.ReadCloser) (io.ReadCloser, error) {
 	return &lz4Reader{
 		src: src,
@@ -96,6 +103,7 @@ func (l *lz4Compressor) Reader(src io.ReadCloser) (io.ReadCloser, error) {
 	}, nil
 }
 
+// Writer returns io.WriteCloser implementation.
 func (l *lz4Compressor) Writer(dst io.WriteCloser) (io.WriteCloser, error) {
 	return &lz4Writer{
 		dst: dst,
@@ -108,10 +116,13 @@ type lz4Reader struct {
 	r   io.Reader
 }
 
+// Read returns the number of bytes for read p (0 <= n <= len(p)).
+// If any errors occurs, it will return an error.
 func (l *lz4Reader) Read(p []byte) (n int, err error) {
 	return l.r.Read(p)
 }
 
+// Close closes src and r.
 func (l *lz4Reader) Close() (err error) {
 	return l.src.Close()
 }
@@ -121,10 +132,13 @@ type lz4Writer struct {
 	w   io.WriteCloser
 }
 
+// Write returns the number of bytes written from p (0 <= n <= len(p)).
+// If any errors occurs, it will return an error.
 func (l *lz4Writer) Write(p []byte) (n int, err error) {
 	return l.w.Write(p)
 }
 
+// Close closes dst and w.
 func (l *lz4Writer) Close() (err error) {
 	err = l.w.Close()
 	if err != nil {
--- a/internal/compress/lz4.go
+++ b/internal/compress/lz4.go
@@ -122,7 +122,7 @@ func (l *lz4Reader) Read(p []byte) (n int, err error) {
 	return l.r.Read(p)
 }
 
-// Close closes src and r.
+// Close closes the reader.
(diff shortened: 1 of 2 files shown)


internal/compress/lz4.go
```diff
@@ -121,10 +132,13 @@ type lz4Writer struct {
 	w   io.WriteCloser
 }
 
+// Write returns the number of bytes written from p (0 <= n <= len(p)).
+// If any errors occurs, it will return an error.
 func (l *lz4Writer) Write(p []byte) (n int, err error) {
 	return l.w.Write(p)
 }
 
+// Close closes dst and w.
```
kevindiu, line 141:
```suggestion
// Close closes the writer.
```

internal/compress/lz4.go
```diff
@@ -108,10 +116,13 @@ type lz4Reader struct {
 	r   io.Reader
 }
 
+// Read returns the number of bytes for read p (0 <= n <= len(p)).
+// If any errors occurs, it will return an error.
 func (l *lz4Reader) Read(p []byte) (n int, err error) {
 	return l.r.Read(p)
 }
 
+// Close closes src and r.
```
kevindiu, line 125:
```suggestion
// Close closes the reader.
```

internal/compress/lz4.go
```diff
@@ -108,10 +116,13 @@ type lz4Reader struct {
...
 func (l *lz4Reader) Close() (err error) {
 	return l.src.Close()
```
kevindiu, line 127: maybe we forget to close `l.r`?

internal/compress/lz4.go
```diff
@@ -121,10 +132,13 @@ type lz4Writer struct {
...
+// Write returns the number of bytes written from p (0 <= n <= len(p)).
+// If any errors occurs, it will return an error.
 func (l *lz4Writer) Write(p []byte) (n int, err error) {
 	return l.w.Write(p)
 }
 
+// Close closes dst and w.
 func (l *lz4Writer) Close() (err error) {
 	err = l.w.Close()
 	if err != nil {
```
kevindiu, line 144: I am not sure should we return err here. If error occurs the `dst` will not be closed.

internal/compress/lz4_test.go
```diff
@@ -91,31 +53,36 @@ func TestNewLZ4(t *testing.T) {
...
+						gob, _ = NewGob()
+						return
+					}(),
+					compressionLevel: 0,
+					lz4:              lz4.New(),
+				},
+			},
+		},
+		{
+			name: "returns (nil, error) when option apply fails",
```
kevindiu, line 73: I think we need `returns (Compressor, nil) when option apply success` test case

vankichi: the first test pattern covers your suggestion.
it has `defaultLZ4Opts` and apply it.

internal/compress/lz4.go
vankichi: Both of `dst` and `w` returns error.

internal/compress/lz4.go
vankichi: `r` is `io.Reader` not `io.ReadCloser`.
we do not have to care about it.

internal/compress/lz4_test.go
kevindiu: But I think it misses the test case to set the custom option 🤔

internal/compress/lz4.go
kevindiu: I mean in line 145, if `l.w.Close()` returns error, line 148 will not executed and I am not sure if it is correct implementation.

internal/compress/lz4_test.go
vankichi: well...
OK, I'll try it.

internal/compress/lz4.go
vankichi: ```return errors.Wrap(l.dst.Close(), err.Error())```
It seems no problem. 🤔 
error of `l.w.Close()`will be wrapped with error of `l.dst.Close()`.

internal/compress/lz4_test.go
```diff
@@ -171,39 +139,162 @@ func Test_lz4Compressor_CompressVector(t *testing.T) {
...
+							WriteFunc: w.Write,
+							FlushFunc: func() error {
+								return nil
+							},
+							CloseFunc: func() error {
+								return nil
+							},
+						}
+					},
+				},
```
hlts2, line 214:
```suggestion
```

internal/compress/lz4_test.go
```diff
@@ -171,39 +139,162 @@ func Test_lz4Compressor_CompressVector(t *testing.T) {
...
+				compressionLevel: 0,
+				lz4: &lz4.MockLZ4{
+					NewWriterLevelFunc: func(w io.Writer, level int) lz4.Writer {
+						return &lz4.MockWriter{
+							WriteFunc: func(p []byte) (int, error) {
+								return 0, errors.New("write err")
+							},
+							FlushFunc: func() error {
+								return nil
+							},
```
hlts2, line 241:
```suggestion
```

internal/compress/lz4_test.go
```diff
@@ -171,39 +139,162 @@ func Test_lz4Compressor_CompressVector(t *testing.T) {
...
+								return errors.New("close err")
+							},
+						}
+					},
+				},
+			},
+			want: want{
+				want: nil,
+				err:  errors.New("close err"),
+			},
```
hlts2, line 314: Could you please add test case that occurs error when Write or Flush and Clouse returns error ?

internal/compress/lz4_test.go
```diff
@@ -793,33 +877,42 @@ func Test_lz4Writer_Close(t *testing.T) {
...
+					},
+				},
+			},
+			want: want{
+				err: errors.New("close err"),
+			},
+		},
 	}
 
 	for _, test := range tests {
```
hlts2, line 936: Could you please add test code for E2E ? (E2E dose not use mock)
for example is this.
https://github.com/vdaas/vald/blob/dcc3a7c3faf1e045170e0300279044fd4f8db9fd/internal/compress/gzip_test.go#L303

hlts2 (changes requested):

internal/compress/lz4_test.go
vankichi: ??? I already created.
what did you mean?

internal/compress/lz4_test.go
vankichi: I'm not sure about that.
when I run `gotests` command, this test will not be created.

internal/compress/lz4_test.go
hlts2: sorry. :bow:
I could not find the following test case...

```go
{
			name: "returns (nil, error) when Close fails",
			args: args{
				vector: []float32{0.1, 0.2, 0.3},
			},
			fields: fields{
				gobc: &MockCompressor{
					CompressVectorFunc: func(vector []float32) (bytes []byte, err error) {
						return []byte("vdaas/vald"), nil
					},
				},
				compressionLevel: 0,
				lz4: &lz4.MockLZ4{
					NewWriterLevelFunc: func(w io.Writer, level int) lz4.Writer {
						return &lz4.MockWriter{
							WriteFunc: func(p []byte) (int, error) {
								return 0, errors.New("write err")
							},
							FlushFunc: func() error {
								return nil
							},
							CloseFunc: func() error {
								return errors.New("close err")
							},
						}
					},
				},
			},
			want: want{
				want: nil,
				err:  errors.Wrap(errors.New("write err"),errors.New("close err").Error())
			},
		},
```

but you have no need to fix it, as it has coverage.
I was making a mistake.:bow:

hlts2 (dismissed):
LGTM

kevindiu (approved):
LGTM

PR #738, opened 2020-10-01
Add test case for s3

hlts2:

### Description:

I added the test case for `s3.go`
And I refactored.

#### Refactor

1. I am writing test about s3/s3.go now. But There were some places where I couldn’t write a test.

- WHY
The current implementation calls New method  inside test target function.
https://github.com/vdaas/vald/blob/master/internal/db/storage/blob/s3/s3.go#L68-L92
So I want to change it so that it can be mocked and tested.

- WHAT
I want to create io interface to create object via interafce like comress. (https://github.com/vdaas/vald/blob/master/internal/compress/lz4/lz4.go#L49)

```go
func (c *client) Reader(ctx context.Context, key string) (io.ReadCloser, error) {
	r := c.readWriter.NewReader(  <= this one
		.....
	)
	return r, r.Open(ctx)
}
***********hole example**********
type readWriter interface {
    NewReader...
    NewWriter ...
}
type client struct {
	eg      errgroup.Group
	readWriter readWriter
}
```

2. Panic occurs in certain cases

- WHY
when option is empty, panic error occurs. because s3 session is nil.
https://github.com/vdaas/vald/blob/master/internal/db/storage/blob/s3/s3.go#L55

- WHAT
I want to nil check for s3 session.
and if session is nil, New returns error.

```go
if c.session == nil {
    return nil, errors.S3SessionNotFound
}
c.service = s3.New(c.session)

```
### Environment:

- Go Version: 1.15.2
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

- [x] I have updated the documentation accordingly.

### DIFF
--- a/internal/db/storage/blob/s3/s3.go
+++ b/internal/db/storage/blob/s3/s3.go
@@ -31,6 +31,25 @@ import (
 	"github.com/vdaas/vald/internal/errors"
 )
 
+type readWriter interface {
+	NewReader(opts ...reader.Option) reader.Reader
+	NewWriter(opts ...writer.Option) writer.Writer
+}
+
+type rw struct{}
+
+func newRW() readWriter {
+	return new(rw)
+}
+
+func (*rw) NewReader(opts ...reader.Option) reader.Reader {
+	return reader.New(opts...)
+}
+
+func (*rw) NewWriter(opts ...writer.Option) writer.Writer {
+	return writer.New(opts...)
+}
+
 type client struct {
 	eg      errgroup.Group
 	session *session.Session
@@ -42,10 +61,15 @@ type client struct {
 
 	readerBackoffEnabled bool
 	readerBackoffOpts    []backoff.Option
+
+	readWriter readWriter
 }
 
+// New returns Bucket implementation.
 func New(opts ...Option) (blob.Bucket, error) {
-	c := new(client)
+	c := &client{
+		readWriter: newRW(),
+	}
 	for _, opt := range append(defaultOpts, opts...) {
 		if err := opt(c); err != nil {
 			return nil, errors.ErrOptionFailed(err, reflect.ValueOf(opt))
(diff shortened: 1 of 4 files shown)


internal/db/storage/blob/s3/s3.go
```diff
@@ -42,31 +61,43 @@ type client struct {
...
+	if c.session == nil {
+		return nil, errors.ErrS3SessionNotFound
+	}
+
 	c.service = s3.New(c.session)
 
 	return c, nil
 }
 
+// Open method returns an error to align the interface, but it doesn't actually return an error.
```
kevindiu, line 88: Open do nothing

internal/db/storage/blob/s3/s3.go
```diff
@@ -42,31 +61,43 @@ type client struct {
...
 func (c *client) Open(ctx context.Context) (err error) {
 	return nil
 }
 
+// Close method returns an error to align the interface, but it doesn't actually return an error.
```
kevindiu, line 93:
```suggestion
// Close do nothing
```

internal/db/storage/blob/s3/s3.go
```diff
@@ -31,6 +31,25 @@ import (
 	"github.com/vdaas/vald/internal/errors"
 )
 
+type readWriter interface {
```
kevindiu, line 34: Please use another name as it is not reader or writer

internal/db/storage/blob/s3/s3_test.go
```diff
@@ -50,42 +53,91 @@ func TestNew(t *testing.T) {
...
+		} else if w.want == nil && got == nil {
+			return nil
 		}
+
+		wantC, gotC := w.want.(*client), got.(*client)
+
+		clientComparator := []comparator.Option{
+			comparator.AllowUnexported(*gotC),
+			comparator.Comparer(func(x, y *session.Session) bool {
+				return x != nil && y != nil
```
kevindiu, line 68: should we use reflect.Equal here?

hlts2: we can not use reflect.DeepEqual here because session contains function type fields. so reflect.DeepEqual always returns false.

kevindiu: or we can use gocmp to define the custom compartor for the function type

PR #760, opened 2020-10-08
add context argument for backoff & add trace span for backoff

kpango:
Signed-off-by: kpango <i.can.feel.gravity@gmail.com>

### Description:
this PR include 2 feature for backoff
1. add context argument for backoff function, It will reduce cancel duration when backoff is timed out
2. add trace span for backoff

### Environment:

- Go Version: 1.15.2
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/backoff/backoff.go
+++ b/internal/backoff/backoff.go
@@ -48,6 +48,8 @@ type Backoff interface {
 	Close()
 }
 
+const traceTag = "vald/internal/backoff/Backoff.Do/retry"
+
 func New(opts ...Option) Backoff {
 	b := new(backoff)
 	for _, opt := range append(defaultOpts, opts...) {
@@ -68,7 +70,7 @@ func (b *backoff) Do(ctx context.Context, f func(ctx context.Context) (val inter
 		return
 	}
 
-	ctx, span := trace.StartSpan(ctx, "vald/internal/backoff/Backoff.Do/retry")
+	ctx, span := trace.StartSpan(ctx, traceTag)
 	defer func() {
 		if span != nil {
 			span.End()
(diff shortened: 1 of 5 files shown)


internal/backoff/backoff.go
```diff
@@ -60,12 +62,18 @@ func New(opts ...Option) Backoff {
...
 	if err == nil || !ret {
 		return
 	}
 
+	ctx, span := trace.StartSpan(ctx, "vald/internal/backoff/Backoff.Do")
+	defer func() {
+		if span != nil {
+			span.End()
+		}
+	}()
```
rinx, line 76: I think this block should be placed before L66.

kpango: Exactly, I'm considered about it, but first execution is not backoff, do we need to trace first execution?

rinx: in the current impl, execution time of retries can be traced by `vald/internal/backoff/Backoff.Do/${cnt}` spans but the execution time of the first one cannot be traced. if we include the first one in the `vald/internal/backoff/Backoff.Do` span, we can calculate it from the traces.

if we adopt the current impl, the name of the span should be renamed as `vald/internal/backoff/Backoff.Do/backoff-executions` because the name `vald/internal/backoff/Backoff.Do` leads misunderstanding that it covers the function `Do` entirely.

kpango: If we trace a function, we should implement the trace code inside the function, so we don't need to trace the execution time of the first function, and it is more efficient to trace only the case of retry execution so that the nesting of the trace does not become deep, so I'm going to rename trace tags but `vald/internal/backoff/Backoff.Do/backoff-executions` sounds redundant how about `vald/internal/backoff/Backoff.Do/retry` or `vald/internal/backoff/Backoff.Do/execute`

rinx: okay.
`vald/internal/backoff/Backoff.Do/retry` and `vald/internal/backoff/Backoff.Do/retry/${i}` sounds better.

kpango: ok! I'll fix ti

kpango:
@rinx fixed

rinx:
Thanks! looks good. :+1:

PR #822, opened 2020-11-09
Add reconciliation loop for job resource

hlts2:
### Description:

we (@kevindiu @hlts2 ) added reconciliation loop logic for job resource.

### Environment:

- Go Version: 1.15.2
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/k8s/job/job.go
+++ b/internal/k8s/job/job.go
@@ -70,7 +70,7 @@ func (r *reconciler) Reconcile(req reconcile.Request) (res reconcile.Result, err
 		return
 	}
 
-	jobs := make(map[string][]Job, len(js.Items))
+	jobs := make(map[string][]Job)
 
 	for _, job := range js.Items {
 		name, ok := job.GetObjectMeta().GetLabels()["app"]
@@ -79,12 +79,16 @@ func (r *reconciler) Reconcile(req reconcile.Request) (res reconcile.Result, err
 			name = strings.Join(jns[:len(jns)-1], "-")
 		}
 
-		// TODO: pre-alocate job slice.
 		if _, ok := jobs[name]; !ok {
 			jobs[name] = make([]Job, 0, len(js.Items))
 		}
 
 		jobs[name] = append(jobs[name], job)
+	}
+
+	for name := range jobs {
+		l := len(jobs[name])
+		jobs[name] = jobs[name][:l:l]
 	}
 
 	if r.onReconcile != nil {
--- a/internal/k8s/job/job.go
+++ b/internal/k8s/job/job.go
@@ -18,6 +18,9 @@ import (
 	"github.com/vdaas/vald/internal/log"
 )
 
+// JobWatcher typealias for k8s.ResourceController.
+type JobWatcher k8s.ResourceController
+
 type reconciler struct {
 	ctx         context.Context
 	mgr         manager.Manager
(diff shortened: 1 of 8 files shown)


kevindiu (approved):
LGTM

kevindiu (approved):

kevindiu (approved):

internal/k8s/metrics/node/node.go
```diff
@@ -61,7 +61,7 @@ func New(opts ...Option) NodeWatcher {
 }
 
 func (r *reconciler) Reconcile(req reconcile.Request) (res reconcile.Result, err error) {
-	m := &metrics.NodeMetricsList{}
+	m := new(metrics.NodeMetricsList)
```
kmrmt, line 64: It is not essential, but I think this fix and the similar ones should be in the other PR.
How about it?

internal/k8s/job/job.go
```diff
@@ -0,0 +1,143 @@
...
+	mgr         manager.Manager
+	name        string
+	namespance  string
+	onError     func(err error)
+	onReconcile func(jobList map[string][]Job)
+}
+
+// Job represents k8s job imformation
+// https://godoc.org/k8s.io/api/batch/v1#Job
+type Job struct {
```
rinx, line 36: Why it is not a type alias for `k8s.io/api/batch/v1.Job`? it is better to use that one if this package is an internal package that is used generally in various applications.

kevindiu: @kpango what do you think about this?

kpango: In the Reconcile loop of the pod package, I had to calculate the CPU/memory requirements and limits of the pod, so I defined a specific type (not an alias for the type), but in the Job package I don't think we need to calculate it, so we can remove the Job struct and  add the type alias or use the directory reference of `k8s.io/api/batch/v1.Job`.

kevindiu: FYI all of the current implementation of reconciler has the same problem/question 
e.g.
https://github.com/vdaas/vald/blob/master/internal/k8s/node/node.go#L91

kpango: In node package and pod package we need to define specific structure type not type alias.

internal/k8s/metrics/node/node.go
hlts2: I think so, too.:bow:
But it's not a big change, so I don't think this change is a problem.
And kpango asked us to include this change in the same PR.

internal/k8s/job/job.go
```diff
@@ -0,0 +1,132 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/vdaas/vald/internal/errors"
+	"github.com/vdaas/vald/internal/k8s"
+	"github.com/vdaas/vald/internal/log"
+)
+
+// JobWatcher is is a type alias for k8s resource controller.
```
kevindiu, line 22:
```suggestion
// JobWatcher is a type alias for k8s resource controller.
```

internal/k8s/job/job.go
```diff
@@ -0,0 +1,132 @@
...
+				Requeue:      true,
+				RequeueAfter: time.Second,
+			}, nil
+		}
+		return
+	}
+
+	var (
+		jobs = make(map[string][]Job, len(js.Items))
+	)
```
kevindiu, line 75:
```suggestion
	jobs := make(map[string][]Job, len(js.Items))
```

kevindiu (approved):
LGTM

vankichi (approved):

PR #823, opened 2020-11-09
Add replicaset reconciler

kevindiu:
### Description:

This PR implements replicaset reconciler for rebalancer.
It implements [k8s reconciler interface](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler) to get the replica set information from k8s API and execute hook when the reconciliation is executed.

It is a part of the rebalancer implementation where the rebalancer will determine wether to execute Vald vector rebalance logic base on the replica set changes on k8s.

### Environment:

- Go Version: 1.15.2
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/k8s/replicaset/option.go
+++ b/internal/k8s/replicaset/option.go
@@ -1,3 +1,18 @@
+//
+// Copyright (C) 2019-2020 Vdaas.org Vald team ( kpango, rinx, kmrmt )
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
 package replicaset
 
 import "sigs.k8s.io/controller-runtime/pkg/manager"
--- a/internal/k8s/replicaset/option.go
+++ b/internal/k8s/replicaset/option.go
@@ -1,0 +1,40 @@
+package main
+
+import "sigs.k8s.io/controller-runtime/pkg/manager"
+
+// Option represents functional option for reconciler.
+type Option func(*reconciler) error
+
+var defaultOpts = []Option{}
+
+// WithControllerName returns Option that sets r.name.
+func WithControllerName(name string) Option {
+	return func(r *reconciler) error {
+		r.name = name
+		return nil
+	}
+}
+
+// WithManager returns Option that sets r.mgr.
+func WithManager(mgr manager.Manager) Option {
+	return func(r *reconciler) error {
+		r.mgr = mgr
+		return nil
+	}
+}
+
+// WithOnErrorFunc returns Option that sets r.onError.
+func WithOnErrorFunc(f func(err error)) Option {
+	return func(r *reconciler) error {
+		r.onError = f
+		return nil
+	}
+}
+
+// WithOnReconcileFunc returns Option that sets r.onReconcile.
+func WithOnReconcileFunc(f func(replicasetList map[string][]ReplicaSet)) Option {
+	return func(r *reconciler) error {
+		r.onReconcile = f
+		return nil
+	}
+}
--- a/internal/k8s/replicaset/option.go
+++ b/internal/k8s/replicaset/option.go
@@ -1,4 +1,4 @@
-package main
+package replicaset
 
 import "sigs.k8s.io/controller-runtime/pkg/manager"
 
@@ -7,7 +7,7 @@ type Option func(*reconciler) error
 
 var defaultOpts = []Option{}
 
-// WithControllerName returns Option that sets r.name.
+// WithControllerName returns Option to set the controller name.
 func WithControllerName(name string) Option {
 	return func(r *reconciler) error {
 		r.name = name
@@ -15,7 +15,15 @@ func WithControllerName(name string) Option {
 	}
 }
 
-// WithManager returns Option that sets r.mgr.
+// WithNamespace returns Option to set the namespace.
+func WithNamespace(ns string) Option {
+	return func(r *reconciler) error {
+		r.namespace = ns
+		return nil
+	}
+}
+
+// WithManager returns Option to set the resource manager.
 func WithManager(mgr manager.Manager) Option {
 	return func(r *reconciler) error {
 		r.mgr = mgr
@@ -23,7 +31,7 @@ func WithManager(mgr manager.Manager) Option {
 	}
 }
 
-// WithOnErrorFunc returns Option that sets r.onError.
+// WithOnErrorFunc returns Option to set the onError hook.
 func WithOnErrorFunc(f func(err error)) Option {
 	return func(r *reconciler) error {
 		r.onError = f
@@ -31,7 +39,7 @@ func WithOnErrorFunc(f func(err error)) Option {
 	}
 }
 
-// WithOnReconcileFunc returns Option that sets r.onReconcile.
+// WithOnReconcileFunc returns Option to set the onReconcile hook.
 func WithOnReconcileFunc(f func(replicasetList map[string][]ReplicaSet)) Option {
 	return func(r *reconciler) error {
 		r.onReconcile = f
--- a/internal/k8s/replicaset/option.go
+++ b/internal/k8s/replicaset/option.go
@@ -23,7 +23,7 @@ func WithNamespace(ns string) Option {
 	}
 }
 
-// WithManager returns Option to set the resource manager.
+// WithManager returns Option to set the controller manager.
 func WithManager(mgr manager.Manager) Option {
 	return func(r *reconciler) error {
 		r.mgr = mgr
--- a/internal/k8s/replicaset/replicaset.go
+++ b/internal/k8s/replicaset/replicaset.go
@@ -1,0 +1,120 @@
+package main
+
+import (
+	"context"
+	"reflect"
+	"strings"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+type ReplicaSetWatcher k8s.ResourceController
+
+type reconciler struct {
+	ctx         context.Context
+	mgr         manager.Manager
+	name        string
+	namespace   string
+	onError     func(err error)
+	onReconcile func(rs []ReplicaSet)
+}
+
+type ReplicaSet struct {
+	Name            string
+	Namespace       string
+	DesiredReplicas *int32
+	Replicas        int32
+}
+
+func New(opts ...Option) (ReplicaSetWatcher, error) {
+	r := new(reconciler)
+
+	for _, opt := range opts {
+		if err := opt(r); err != nil {
+			return erros.ErrOptionFailed(err, reflect.ValueOf(opt))
+		}
+	}
+
+	return r
+}
+
+func (r *reconciler) Reconcile(req reconcile.Request) (res reconcile.Result, err error) {
+	rsl := new(appsv1.ReplicaSetList)
+
+	err = r.mgr.GetClient().List(r.ctx, rsl)
+	if err != nil {
+		if r.onError != nil {
+			r.onError(err)
+		}
+		res = reconcile.Result{
+			Requeue:      true,
+			RequeueAfter: time.Millisecond * 100,
+		}
+		if errors.IsNotFound(err) {
+			res = reconcile.Result{
+				Requeue:      true,
+				RequeueAfter: time.Second,
+			}
+		}
+		return
+	}
+
+	rs := make(map[string][]ReplicaSet, len(rsl.Items))
+
+	for _, replicaset := range rsl.Items {
+		name, ok := replicaset.GetObjectMeta().GetLabels()["app"]
+		if !ok {
+			pns := strings.Split(pod.GetName(), "-")
+			name = strings.Join(pns[:len(pns)-1], "-")
+		}
+
+		if _, ok := rs[name]; !ok {
+			rs[name] = make([]ReplicaSet, 0, len(rsl.Items))
+		}
+
+		rs[name] = append(rs[name], ReplicaSet{
+			Name:            name,
+			Namespace:       replicaset.GetNamespace(),
+			DesiredReplicas: replicaset.ReplicaSetSpec.Replicas,
+			Replicas:        replicaset.Status.Replicas,
+		})
+	}
+
+	if r.onReconcile != nil {
+		r.onReconcile(rs)
+	}
+
+	return
+}
+
+func (r *reconciler) GetName() string {
+	return r.name
+}
+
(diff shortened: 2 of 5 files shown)


internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,120 @@
...
+type reconciler struct {
+	ctx         context.Context
+	mgr         manager.Manager
+	name        string
+	namespace   string
+	onError     func(err error)
+	onReconcile func(rs []ReplicaSet)
+}
+
+type ReplicaSet struct {
```
kevindiu, line 26: same as https://github.com/vdaas/vald/pull/822/files#r519616815

hlts2 (approved):
LGTM

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,127 @@
...
+	name        string
+	namespace   string
+	onError     func(err error)
+	onReconcile func(rs map[string][]ReplicaSet)
+}
+
+// ReplicaSet is a type alias for the k8s replica set definition.
+type ReplicaSet = appsv1.ReplicaSet
+
+// New returns the replica set watcher that implements reconciler loop, or any error occurred.
```
hlts2, line 36:
```suggestion
// New returns the ReplicaSetWatcher that implements reconciler loop, or any error occurred.
```

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, opt := range opts {
+		if err := opt(r); err != nil {
+			return nil, errors.ErrOptionFailed(err, reflect.ValueOf(opt))
+		}
+	}
+
+	return r, nil
+}
+
+// Reconcile implements k8s reconciler loop to retrieve the replica set information from k8s.
```
hlts2, line 49:
```suggestion
// Reconcile implements k8s reconciliation loop to retrieve the replica set information from k8s.
```

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return
+}
+
+// GetName returns the name of the replica set watcher.
+func (r *reconciler) GetName() string {
+	return r.name
+}
+
+// NewReconciler returns the reconciler for the replica set watcher.
```
hlts2, line 99:
```suggestion
// NewReconciler returns the reconciler for the ReplicaSet.
```

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,127 @@
...
+		r.ctx = ctx
+	}
+	if r.mgr == nil && mgr != nil {
+		r.mgr = mgr
+	}
+	appsv1.AddToScheme(r.mgr.GetScheme())
+	return r
+}
+
+// For returns the reconcile object type which is replica set.
```
hlts2, line 111:
```suggestion
// For returns the runtime.Object which is replica set.
```

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,127 @@
...
+	name        string
+	namespace   string
+	onError     func(err error)
+	onReconcile func(rs map[string][]ReplicaSet)
+}
+
+// ReplicaSet is a type alias for the k8s replica set definition.
+type ReplicaSet = appsv1.ReplicaSet
+
+// New returns the ReplicaSetWatcher that implements reconciler loop, or any error occurred.
```
hlts2, line 36:
```suggestion
// New returns the ReplicaSetWatcher that implements reconciliation loop, or any error occurred.
```

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	if r.onReconcile != nil {
+		r.onReconcile(rs)
+	}
+
+	return
+}
+
+// GetName returns the name of the replica set watcher.
```
hlts2, line 94:
```suggestion
// GetName returns the name of resource controller.
```

internal/k8s/replicaset/option.go
```diff
@@ -0,0 +1,48 @@
...
+
+// WithNamespace returns Option to set the namespace.
+func WithNamespace(ns string) Option {
+	return func(r *reconciler) error {
+		r.namespace = ns
+		return nil
+	}
+}
+
+// WithManager returns Option to set the resource manager.
```
hlts2, line 26:
```suggestion
// WithManager returns Option to set the controller manager.
```

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,127 @@
...
+	return r
+}
+
+// For returns the runtime.Object which is replica set.
+func (r *reconciler) For() runtime.Object {
+	return new(appsv1.ReplicaSet)
+}
+
+// Owns return the owner of the replica set watcher.
+// It will always returns nil.
```
vankichi, line 117:
```suggestion
// Owns returns the owner of the replica set watcher.
// It will always return nil.
```

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,127 @@
...
+func (r *reconciler) Owns() runtime.Object {
+	return nil
+}
+
+// Watches returns the kind of the replica set and the event handler.
+// It will always returns nil.
```
vankichi, line 123:
```suggestion
// Watches returns the kind of the replica set and the event handler.
// It will always return nil.
```

vankichi:
@kevindiu could you add the test files corresponding to the new files you added?

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+		res = reconcile.Result{
+			Requeue:      true,
+			RequeueAfter: time.Millisecond * 100,
+		}
+		if k8serrors.IsNotFound(err) {
+			res = reconcile.Result{
+				Requeue:      true,
+				RequeueAfter: time.Second,
+			}
```
vankichi, line 81:
```suggestion
			res.RequeueAfter = time.Second
```

hlts2 (approved):
LGTM

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,142 @@
...
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/vdaas/vald/internal/errors"
+	"github.com/vdaas/vald/internal/k8s"
+	"github.com/wdiu/k8s-athenz-identity/pkg/log"
```
vankichi, line 34:
```suggestion
```

hlts2 (approved):
LGTM

vankichi (approved):

vankichi:
@kpango @rinx 
could you please review it? it needs no deadline, but we appreciate if you review it soon. 🙏

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		if r.onError != nil {
+			r.onError(err)
+		}
+		res = reconcile.Result{
+			Requeue:      true,
+			RequeueAfter: time.Millisecond * 100,
+		}
+		if k8serrors.IsNotFound(err) {
+			log.Error("not found", err)
```
rinx, line 79: it may be better to use log.xxxf function.
```suggestion
			log.Errorf("not found: %s", err)
```

Is it really ERROR level log? I think it is not so serious situation, just WARN level or lower one.
https://github.com/vdaas/vald/blob/master/docs/contributing/coding-style.md#logging

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,142 @@
...
+			pns := strings.Split(replicaset.GetName(), "-")
+			name = strings.Join(pns[:len(pns)-1], "-")
+		}
+
+		if _, ok := rs[name]; !ok {
+			rs[name] = make([]ReplicaSet, 0, len(rsl.Items))
+		}
+
+		rs[name] = append(rs[name], replicaset)
+	}
```
rinx, line 100: i think it is not so meaningful to put them into a map here.
Do you have a vision to use them?
i think it is enough to use just a slice.
```go
	onReconcile func(rs []ReplicaSet)
```

internal/k8s/replicaset/replicaset.go
kpango: I think it's better to define new error variable in errors package eg. `errors.ErrK8SResourceNotFound(err)`

internal/k8s/replicaset/replicaset.go
vankichi: @rinx 
If we use `[]Replicaset`, we have to loop each app for getting Metadata.
On the other hand, when we use `map[string][]Replicaset`, we can access easily each app's Metadata and do not have to loop for getting it.

rinx: > If we use []Replicaset, we have to loop each app for getting Metadata.

already you have a loop to get `metadata.labels` in L88-L100 even if you use `map[string][]ReplicaSet` type for the argument of `onReconcile`.

even if you use `map[string][]ReplicaSet` type, i think you may have a loop for `[]ReplicaSet` in the `onReconcile` function like:
```go
func onReconcile(rs map[string][]ReplicaSet) {
    xrs, ok := rs["hoge"]
    // ... omission ...
    for _, r := range xrs {
        // do something
    }
}
```
so the loop for `[]ReplicaSet` may duplicate.

This is just a comment. so you don't have to revise it, but i would write it without map.

vankichi: thx, and we just decided use this style.
But, we should consider the performance later, we will consider about it again when simulating performance.

kpango: As you know the k8s/pod,node package uses the same logic, but since the Reconciler can retrieve resources from the entire namespace, you'll need to turn the long loop and do a string comparison of if == when retrieving a Resource with a specific name.
Therefore, by using the fact that the time axis of Reconcile and the time axis for processing Cache are different, it is possible to process only a list of specific resource names in the onReconcile function by grouping resources into a map for each resource name at Reconcile.

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,152 @@
...
+type ReplicaSetWatcher k8s.ResourceController
+
+type reconciler struct {
+	ctx                 context.Context
+	mgr                 manager.Manager
+	name                string
+	namespace           string
+	onError             func(err error)
+	onReconcile         func(rs map[string][]ReplicaSet)
+	lastReconcileResult map[string][]ReplicaSet
```
hlts2, line 47:
```suggestion
	lastReconciledResult map[string][]ReplicaSet
```

hlts2 (approved):

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,152 @@
...
+			pns := strings.Split(replicaset.GetName(), "-")
+			name = strings.Join(pns[:len(pns)-1], "-")
+		}
+
+		if _, ok := rs[name]; !ok {
+			lrr, ok := r.lastReconciledResult[name]
+			if !ok {
+				rs[name] = make([]ReplicaSet, 0, 0)
+			} else {
+				rs[name] = make([]ReplicaSet, 0, len(lrr))
```
kpango, line 103: this code will lost appended data from L#107

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,152 @@
...
+			}
+		}
+
+		rs[name] = append(rs[name], replicaset)
+	}
+
+	r.lastReconciledResult = rs
+
+	if r.onReconcile != nil {
+		r.onReconcile(rs)
```
kpango, line 113: how do you think about race condition, if onReconcile function includes goroutine which makes race condition problem

internal/k8s/replicaset/replicaset.go
kpango: oh missread, I think it's okay but it's not effective,

kpango: why you need to make all reconciliation time?
I think it's better to change pointer seek to zero offset.

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,152 @@
...
+		name, ok := replicaset.GetObjectMeta().GetLabels()["app"]
+		if !ok {
+			pns := strings.Split(replicaset.GetName(), "-")
+			name = strings.Join(pns[:len(pns)-1], "-")
+		}
+
+		if _, ok := rs[name]; !ok {
+			lrr, ok := r.lastReconciledResult[name]
+			if !ok {
+				rs[name] = make([]ReplicaSet, 0, 0)
```
kpango, line 101: do we need cap variable??

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		if k8serrors.IsNotFound(err) {
+			log.Warn(errors.ErrK8sResourceNotFound(err))
+			res.RequeueAfter = time.Second
+			return res, nil
+		}
+		return
+	}
+
+	rs := make(map[string][]ReplicaSet)
```
kpango, line 89: please try to reuse allocated buffer

internal/k8s/replicaset/replicaset.go
kevindiu: Maybe we can't understand what you mean by `lost appended data`.
In line 98, it checks if the cache is initialized.
If the cache is not initialized,  line 99-103 initializes the empty slice `[]ReplicaSet`.

kpango (changes requested):
I think sync.Pool can resolve these problems, like initialization & race condition.

hlts2 (approved):

internal/k8s/replicaset/replicaset.go
```diff
@@ -0,0 +1,161 @@
...
+			return res, nil
+		}
+		return
+	}
+
+	// reset the last result cache
+	lrr := r.pool.Get().(map[string][]ReplicaSet)
+	for name := range lrr {
+		lrr[name] = lrr[name][:0]
+	}
```
kpango, line 98: this code shoud be line 123

vankichi: you mean put the empty slice into pool, right?

kpango: yes, we should store zero seeked slice, it has enough capacity.
we don't need to allocate next time.

kpango: and your code also initialize after pool.Get, it means same.

vankichi: well...So my understanding is that we do not have to store the current reconcile result in the pool because of no using the data anymore (we need to know only the capacity of slices), is this correct along with your comment?

kpango: Yes, the pool should store zero-seek slices with cap allocations, and there is no need to store the data.

vankichi: thx. I got it clearly 😎

PR #831, opened 2020-11-12
Create internal/net/http/client option test

kevindiu:
### Description:

This PR implements http client option test. 
This PR also includes the following changes:
- [Test related] add the comparator implementation of `transport` to the `client_test.go` 
- [Refactor] Refactor the error returned from the option and handle it in `New()` 

### Environment:

- Go Version: 1.15.2
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/net/http/client/option_test.go
+++ b/internal/net/http/client/option_test.go
@@ -494,6 +494,19 @@ func TestWithMaxIdleConns(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "set conn success with 0 value",
+			args: args{
+				cn: 0,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						MaxIdleConns: 0,
+					},
+				},
+			},
+		},
 	}
 
 	for _, test := range tests {
@@ -562,6 +575,19 @@ func TestWithMaxIdleConnsPerHost(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "set conn per host success with 0 value",
+			args: args{
+				cn: 0,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						MaxIdleConnsPerHost: 0,
+					},
+				},
+			},
+		},
 	}
 
 	for _, test := range tests {
@@ -625,6 +651,19 @@ func TestWithMaxConnsPerHost(t *testing.T) {
 				obj: &T{
 					Transport: &http.Transport{
 						MaxConnsPerHost: 5,
+					},
+				},
+			},
+		},
+		{
+			name: "set conn per host success with 0 value",
+			args: args{
+				cn: 0,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						MaxConnsPerHost: 0,
 					},
 				},
 			},
@@ -1057,6 +1096,19 @@ func TestWithMaxResponseHeaderBytes(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "set max response header byte with 0 value",
+			args: args{
+				bs: 0,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						MaxResponseHeaderBytes: 0,
+					},
+				},
+			},
+		},
 	}
 
 	for _, test := range tests {
@@ -1125,6 +1177,19 @@ func TestWithWriteBufferSize(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "set write buffer size with 0 value",
+			args: args{
+				bs: 0,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						WriteBufferSize: 0,
+					},
+				},
+			},
+		},
 	}
 
 	for _, test := range tests {
@@ -1189,6 +1254,19 @@ func TestWithReadBufferSize(t *testing.T) {
 				obj: &T{
 					Transport: &http.Transport{
 						ReadBufferSize: 5,
+					},
+				},
+			},
+		},
+		{
+			name: "set buffer size success with 0 value",
+			args: args{
+				bs: 0,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						ReadBufferSize: 0,
 					},
 				},
 			},
--- a/internal/net/http/client/option_test.go
+++ b/internal/net/http/client/option_test.go
@@ -164,9 +164,7 @@ func TestWithDialContext(t *testing.T) {
 		}(),
 		{
 			name: "return error when dial context is nil",
-			args: args{
-				dx: nil,
-			},
+			args: args{},
 			want: want{
 				obj: &T{
 					Transport: &http.Transport{},
@@ -255,9 +253,7 @@ func TestWithTLSHandshakeTimeout(t *testing.T) {
 		},
 		{
 			name: "set timeout failed with empty value",
-			args: args{
-				dur: "",
-			},
+			args: args{},
 			want: want{
 				obj: &T{
 					Transport: &http.Transport{},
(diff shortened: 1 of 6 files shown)


hlts2 (dismissed):
LGTM

vankichi:
@kevindiu 
why did you change `client_test.go`?
I think it is not needed in this PR 🤔

kevindiu:
@vankichi I think it is better to write transport comparator in `client_test` not in `option_test`, so I changed client_test implementation

internal/net/http/client/option_test.go
```diff
@@ -22,1945 +22,1378 @@ import (
...
+				cn: 5,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						MaxIdleConns: 5,
+					},
+				},
+			},
+		},
```
vankichi, line 496: I think it'd be better when the case of no value will be created.
```go
{
  name: "set conn success with no input"
  want: want{
    obj: &T{
      Transport: &http.Transport{
        MacIdelConns: 0,
      }
    }
  }
}

File: internal/net/http/client/option_test.go [hunk 2]
```diff
@@ -22,1945 +22,1378 @@ import (
...
+				cn: 5,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						MaxIdleConnsPerHost: 5,
+					},
+				},
+			},
+		},
```
vankichi, line 564: - same as Line496

internal/net/http/client/option_test.go
```diff
@@ -22,1945 +22,1378 @@ import (
...
+				cn: 5,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						MaxConnsPerHost: 5,
+					},
+				},
+			},
+		},
```
vankichi, line 631: - same as Line496

internal/net/http/client/option_test.go
```diff
@@ -22,1945 +22,1378 @@ import (
...
+				obj: &T{
+					Transport: &http.Transport{},
+				},
+				err: errors.NewErrCriticalOption("idleConnTimeout", "dummy", errors.New("invalid timeout value: dummy\t:timeout parse error out put failed: time: invalid duration \"dummy\"")),
+			},
+		},
+		{
+			name: "set timeout failed with empty value",
+			args: args{
+				dur: "",
```
vankichi, line 715: I think it is not needed.

internal/net/http/client/option_test.go
```diff
@@ -22,1945 +22,1378 @@ import (
...
+				obj: &T{
+					Transport: &http.Transport{},
+				},
+				err: errors.NewErrCriticalOption("responseHeaderTimeout", "dummy", errors.New("invalid timeout value: dummy\t:timeout parse error out put failed: time: invalid duration \"dummy\"")),
+			},
+		},
+		{
+			name: "set timeout failed with empty value",
+			args: args{
+				dur: "",
```
vankichi, line 807: I think it is not needed.

internal/net/http/client/option_test.go
```diff
@@ -22,1945 +22,1378 @@ import (
...
+				obj: &T{
+					Transport: &http.Transport{},
+				},
+				err: errors.NewErrCriticalOption("expectContinueTimeout", "dummy", errors.New("invalid timeout value: dummy\t:timeout parse error out put failed: time: invalid duration \"dummy\"")),
+			},
+		},
+		{
+			name: "set timeout failed with empty value",
+			args: args{
+				dur: "",
```
vankichi, line 899: I think it is not needed.

internal/net/http/client/option_test.go
```diff
@@ -22,1945 +22,1378 @@ import (
...
+				bs: 5,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						MaxResponseHeaderBytes: 5,
+					},
+				},
+			},
+		},
```
vankichi, line 1059: - same as Line496

internal/net/http/client/option_test.go
```diff
@@ -22,1945 +22,1378 @@ import (
...
+				bs: 5,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						WriteBufferSize: 5,
+					},
+				},
+			},
+		},
```
vankichi, line 1127: - same as Line496

internal/net/http/client/option_test.go
```diff
@@ -22,1945 +22,1378 @@ import (
...
+				bs: 5,
+			},
+			want: want{
+				obj: &T{
+					Transport: &http.Transport{
+						ReadBufferSize: 5,
+					},
+				},
+			},
+		},
```
vankichi, line 1195: - same as Line496

vankichi:
> @vankichi I think it is better to write transport comparator in `client_test` not in `option_test`, so I changed client_test implementation

@kevindiu 
umm...so, you have to show above implementation in your PR description at least and you had to communicate me and @hlts2 before implementation at first. But, it is not a critical problem.

internal/net/http/client/option_test.go
kevindiu: @vankichi In golang, the default value of `int` is 0, so if we write 0 here, we can not differentiate if the value is set or not, it I think this test case may be meaningless.
What do you think?

internal/net/http/client/option_test.go
kevindiu: @vankichi Do you mean this line `dur: ""` is not need?
I think it is more clear to write even if the value is the default value of the type, what do you think?

internal/net/http/client/option_test.go
vankichi: I think it will be set as 0 when the no value is input is not meaningless and better for test coverage.
I think `set 0 with no input`, which means no logic will be run,  is the one of the test  case.

internal/net/http/client/option_test.go
vankichi: but, as you know, the default value of string is `""`.
I do not care but it is not necessary, also your opinion is right.
It is the decision problem, any test rule in coding guideline?

internal/net/http/client/option_test.go
kevindiu: I will fix it, thank you for your opinion

internal/net/http/client/option_test.go
kevindiu: There are no such rule in the coding guideline.
I am not sure if I am correct, as all we know the default value of string is "" in golang, we can decide to write this definition or not by ourself (not standardize for everyone, but preference for each developer), what do you think?

Even if we decide to use this pattern, I think this rule only applies to test code, so I think it is better not to write it in coding guideline.

vankichi: I think it is clear creating test case with no input test case like as `int` pattern.
And, I think we consider about the unit test practice and create new document for unit testing.
How about it?

kevindiu: > I think it is clear creating test case with no input test case like as int pattern.
I think users need to input the value in option (for example `WithXXX("")`, so I think writing `""` is more realistic and easier to understand.

>  I think we consider about the unit test practice and create new document for unit testing.
I think both ok for me. If you think the code quality of test code is important then I think maybe it is better to create it

vankichi: well, so how about write the default value expressly in each test?

OK. Let's try it, but Not now

hlts2 (dismissed):

vankichi (dismissed):
LGTM

PR #852, opened 2020-11-19
Add access log gRPC interceptor & Update observability features

rinx:
#### Changes
- Add access log gRPC interceptor
    - Add internal/net/grpc/interceptor/server/logging package
    - By appending "AccessLogInterceptor" to `server_config.servers.grpc.server.grpc.interceptors` field in config.yaml, it will be enabled.
```
{"level":"info","ts":1606366367.1742995,"msg":"rpc completed","details":{"grpc":{"kind":"unary","service":"core.Agent","method":"IndexInfo"},"startTime":1606366367.1742105,"latency":0.0000113,"traceID":"bd2375a0b3aa4a2607d042416c56420d"}}
```

- Add zap logger
    - it is better JSON formatting than the existing logging library.

- Update Grafana dashboards
- Add values-observability-demo.yaml to charts/vald

#### Known problem
- when logger=glg and format=raw, it cannot print a valid structured log. please use zap instead or disable AccessLogInterceptor.
```
2020-11-26 04:50:33     [INFO]: rpc completed &{0xc00024cc30 1.6063662337968597e+09 1.45e-05 86c42bd1164bf434f95b3f51f9d5b755 <nil>}
```

### Related Issue:

nothing

nothing

### Environment:

- Go Version: 1.15.2
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


internal/log/zap/zap.go
```diff
@@ -0,0 +1,189 @@
...
+		zap.ErrorOutput(errSink),
+	}
+
+	if l.enableCaller {
+		opts = append(opts, zap.AddCaller())
+	}
+
+	l.logger = zap.New(core, opts...)
+
+	l.sugar = l.logger.Sugar()
```
kpango, line 90: I think we should be define finelizer for logger & sugar.
zap has internal buffering we should call .Sync() method when application exits

kpango: FYI: https://pkg.go.dev/go.uber.org/zap#Logger.Sync
https://pkg.go.dev/go.uber.org/zap#SugaredLogger.Sync

internal/net/grpc/interceptor/server/logging/accesslog.go
```diff
@@ -0,0 +1,145 @@
...
+	"path"
+	"time"
+
+	"github.com/vdaas/vald/internal/log"
+	"github.com/vdaas/vald/internal/observability/trace"
+	"google.golang.org/grpc"
+)
+
+type AccessLogEntity struct {
+	Grpc      *AccessLogGrpcEntity `json:"grpc,omitempty" yaml:"grpc"`
```
kpango, line 31:
```suggestion
	GRPC      *AccessLogGRPCEntity `json:"grpc,omitempty" yaml:"grpc"`
```

internal/net/grpc/interceptor/server/logging/accesslog.go
```diff
@@ -0,0 +1,145 @@
...
+		span := trace.FromContext(ctx)
+		if span != nil {
+			traceID = span.SpanContext().TraceID.String()
+		}
+
+		start := time.Now()
+
+		resp, err = handler(ctx, req)
+
+		latency := float64(time.Since(start)) / float64(time.Second)
```
kpango, line 62: I think it would be better to store time as unix nano second scale.
then we can calculate it on the monitoring query

internal/net/grpc/interceptor/server/logging/accesslog.go
```diff
@@ -0,0 +1,145 @@
...
+		err := handler(srv, ss)
+
+		latency := float64(time.Since(start)) / float64(time.Second)
+		startTime := float64(start.UnixNano()) / float64(time.Second)
+
+		service, method := parseMethod(info.FullMethod)
+
+		entity := &AccessLogEntity{
+			Grpc: &AccessLogGrpcEntity{
+				Kind:    "stream",
```
kpango, line 117: can we define "stream" as a const value?

internal/net/grpc/interceptor/server/logging/accesslog.go
```diff
@@ -0,0 +1,145 @@
...
+
+		if traceID != "" {
+			entity.TraceID = traceID
+		}
+
+		if err != nil {
+			entity.Error = err
+			log.Error("rpc completed", entity)
+		} else {
+			log.Info("rpc completed", entity)
```
kpango, line 85: I think debug is better

internal/net/grpc/interceptor/server/logging/accesslog.go
```diff
@@ -0,0 +1,145 @@
...
+			Latency:   latency,
+		}
+
+		if traceID != "" {
+			entity.TraceID = traceID
+		}
+
+		if err != nil {
+			entity.Error = err
+			log.Error("rpc completed", entity)
```
kpango, line 83: may be you want print entity on error log right?
so, entity print should be %#v print pragma because of entity is pointer type , error log will show thier address only...

kpango (changes requested):
Others LGTM

internal/net/grpc/interceptor/server/logging/accesslog.go
rinx: revised.

internal/net/grpc/interceptor/server/logging/accesslog.go
rinx: revised

internal/net/grpc/interceptor/server/logging/accesslog.go
rinx: We should use `log.Error` instead of `log.Errorf` here because Zap logger prints the second argument as the `"details"` field.
https://pkg.go.dev/go.uber.org/zap#Logger.Info
Please see the description of this PR. There's an example.

So you know, there's a known problem (as described in the description) when using Glg logger.

internal/net/grpc/interceptor/server/logging/accesslog.go
rinx: AccessLogInterceptor is disabled by default, so i think it is okay to use Info level here.
I think it is too many logs when turning on Debug logs (such as Agent's CreateIndex operation log). it is better to put access logs in different level compared to them.

kpango: If there is a lot of access, it will be difficult to get other Info logs according to the number of accesses, so I think this Interceptor is used for Debug, I think Debug is fine.

The Info log should contain more important information, and I think it is too much to output every successful access.

internal/net/grpc/interceptor/server/logging/accesslog.go
kpango: Since the log.Error interface is independent of zap, it should not be implemented to work only with zap.
Even if we specify a standard Go logger here, if the group of arguments is the same interface, the address will be displayed as in Glg.
I think it would be more appropriate to consider a more explicit Print method, but I'm not sure.

internal/net/grpc/interceptor/server/logging/accesslog.go
rinx: okay i'll revise it.

kpango: thank you!

internal/net/grpc/interceptor/server/logging/accesslog.go
rinx: The users of this AccessLogInterceptor will use Loki or something to query the aggregated logs. so there's no problem if there is a lot of accesses.
If access logs is Debug level, it is bigger problem that meaningless Debug level logs require large storage capacity to store them.

Also I think it is correct according to contributing guide. https://github.com/vdaas/vald/blob/master/docs/contributing/coding-style.md#logging

internal/net/grpc/interceptor/server/logging/accesslog.go
rinx: hmm...
I don't know how to print correctly structured JSON logs with that format without using `log.Error`.
Okay, I'll close this PR.

internal/net/grpc/interceptor/server/logging/accesslog.go
kpango: hmm, but I'd like to divide trace log and Info log, so how about make new log level called TRACE?

internal/net/grpc/interceptor/server/logging/accesslog.go
kpango: wait, you don't need to close this pr.
let's discuss about it.

this pr feature is really effective.

PR #909, opened 2021-01-06
fix grpc addrs definition for v1 new design api

kpango:

### Description:
this pr fixes grpc addr setting for V1 API
current implementation (v0.0.65) defines HOST & PORT for each grpc connection, but grpc config also have addrs string slice field this pr removes HOST & PORT field from yaml & configuration & use grpc.addrs instead of them.

### Environment:

- Go Version: 1.15.2
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


kpango:
@rinx this PR is ready for review, I'd appreciate if you could review.

Please ignore the build failure because the version of the controller runtime is different in the base image dependency and therefore the discoverer docker image cannot be built It will work when merging PR-826.

charts/vald/templates/gateway/backup/ing.yaml
```diff
@@ -31,11 +31,10 @@ metadata:
     app.kubernetes.io/component: gateway-backup
   name: {{ $gateway.name }}-ingress
 spec:
+  defaultBackend:
+    service:
+      name: {{ $gateway.name }}
+      {{- include "vald.ingressPort" (dict "Values" $gateway.ingress) | nindent 6 }}
```
rinx, line 37: I think it should not be in the `defaultBackend` object.

> DefaultBackend is the backend that should handle requests that don't match any rule.

I'm not sure but it should be moved into `rules`? Otherwise, maybe we cannot have multiple ingresses in a single cluster.

charts/vald/templates/gateway/lb/ing.yaml
```diff
@@ -31,11 +31,10 @@ metadata:
     app.kubernetes.io/component: gateway-lb
   name: {{ $gateway.name }}-ingress
 spec:
+  defaultBackend:
+    service:
+      name: {{ $gateway.name }}
+      {{- include "vald.ingressPort" (dict "Values" $gateway.ingress) | nindent 6 }}
```
rinx, line 37: same

charts/vald/templates/gateway/meta/ing.yaml
```diff
@@ -31,11 +31,10 @@ metadata:
     app.kubernetes.io/component: gateway-meta
   name: {{ $gateway.name }}-ingress
 spec:
+  defaultBackend:
+    service:
+      name: {{ $gateway.name }}
+      {{- include "vald.ingressPort" (dict "Values" $gateway.ingress) | nindent 6 }}
```
rinx, line 37: same

charts/vald/templates/gateway/vald/ing.yaml
```diff
@@ -31,11 +31,10 @@ metadata:
     app.kubernetes.io/component: gateway
   name: {{ $gateway.name }}-ingress
 spec:
+  defaultBackend:
+    service:
+      name: {{ $gateway.name }}
+      {{- include "vald.ingressPort" (dict "Values" $gateway.ingress) | nindent 6 }}
```
rinx, line 37: same

docs/tutorial/get-started.md
```diff
@@ -55,14 +55,13 @@ If you want to learn about Scylla, please refer to [the official website](https:
     Deploy Scylla as a backup database.
 
     ```bash
-    kubectl apply -f k8s/jobs/db/initialize/cassandra/configmap.yaml
-    kubectl apply -f k8s/external/scylla
+    make k8s/external/scylla/deploy
```
rinx, line 58: Is it okay that users use Makefile tasks? :thinking:

k8s/gateway/vald/ing.yaml
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-apiVersion: networking.k8s.io/v1beta1
+apiVersion: networking.k8s.io/v1
```
rinx, line 17: NOTE:
Because the apiVersion is updated, the `servicePort` field should also be updated.
But it is okay to ignore it because this file will be automatically updated by CI when a new release is created.

charts/vald/templates/gateway/backup/ing.yaml
kpango: I know about the use of DefaultBackend, but when using the original(master branch) rules.http.paths.backend, networking.k8s.io/v1 now requires path and pathType field, and since we didn't use pathbased routing in this case, we're currently using single host In this case, the routing was not used in path, so now we should use DefaultBackend for single-host Ingress.

charts/vald/templates/gateway/lb/ing.yaml
kpango: same

charts/vald/templates/gateway/meta/ing.yaml
kpango: same

charts/vald/templates/gateway/vald/ing.yaml
kpango: same

docs/tutorial/get-started.md
kpango: how do you think?
I think scylla is not core value of this document, imo it's okay.

k8s/gateway/vald/ing.yaml
kpango: I know. we can ignore it this time.

charts/vald/templates/gateway/backup/ing.yaml
rinx: I've just tested on my kubernetes cluster with Traefik Ingress controller, it doesn't work when only using defaultBackend like this branch. 
It works when using a yaml like below.
```yaml
spec:
  rules:
  - host: xxx.yyy.zzz #dummy
    http:
      paths:
      - backend:
          service:
            name: vald-gateway
            port:
              name: grpc
        pathType: ImplementationSpecific
```
As a result, at least the case of Traefik Ingress controller, pathType = ImplementationSpecific without defaultBackend is correct configuration. The chart should be revised. (Sorry I don't know about other Ingress controller)

docs/tutorial/get-started.md
rinx: I think the users who are reading Get Started document should know what is done in the example script. So it is better to write **exact** commands in the document. I think the former document is better because Makefile hides the exact commands.

charts/vald/templates/gateway/backup/ing.yaml
kpango: In my envitonment it's works...
I think it's coused on Traefik ??
https://kubernetes.io/docs/concepts/services-networking/ingress/#types-of-ingress

How do you think about pathType, ImplementationSpecific is not common, but we use gRPC each gateway has same prefix, I think it would be better to set Exact matching for PathType.

rinx: it's confusing... :sweat_smile:

rinx: are you using NGINX ingress controller?

kpango: yes, and I tried contour....

kpango: but Vald should be work on any environment.
we should fix it to the rules base routing.

kpango: I'm not sure which vhost should we set on Ingress?

kpango: as you defined as dummy.

rinx: i set the VM host of the corp internal VM to vhost.

kpango: you mean specific node's host??
or dns name??

rinx: my cluster is a single node cluster, so i set the DNS name of the node host in my test.

docs/tutorial/get-started.md
kpango: should we describe about scylla's exact commands??

docs/tutorial/get-started.md
```diff
@@ -55,14 +55,13 @@ If you want to learn about Scylla, please refer to [the official website](https:
...
     ```
 
     Apply kubernetes metrics-server
 
     ```bash
-    kubectl apply -f k8s/metrics/metrics-server
+    make k8s/metrics/metrics-server/deploy
```
kpango, line 64: metrics-server deploy command is simple, as you [mentioned](https://github.com/vdaas/vald/pull/909/files/82fd3963055b909cdf833b65317d4c09603cba99?file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.mk&file-filters%5B%5D=.tpl&file-filters%5B%5D=.yaml#r553701026) we can write kubectl apply.

charts/vald/templates/gateway/backup/ing.yaml
kpango: fixed thx!

rinx (approved):

docs/tutorial/get-started.md
kpango: I added more description about this section, thank you for your notice.

PR #944, opened 2021-01-21
Add Zap logger and access log interceptor

rinx:
### Description:

This PR is cherry-picked from #852 and revised.
- Add Zap logger.
- Revise logger interfaces to add Debugd, Infod, etc...
- Add logger finalizer method `Close()`
- Add AccessLogInterceptor

### Related Issue:

nothing

nothing

### Environment:

- Go Version: 1.15.7
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.13.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


internal/log/glg/glg.go
```diff
@@ -81,6 +85,10 @@ func (l *logger) Infof(format string, vals ...interface{}) {
 	l.retry.Outf(l.glg.Infof, format, vals...)
 }
 
+func (l *logger) Infod(msg string, details ...interface{}) {
+	l.Infof(msg+"\tdetails: %#v", details)
```
rinx, line 89: Is this format okay? :thinking:

kpango: please define default details format as const value on the top

internal/log/zap/zap.go
```diff
@@ -0,0 +1,222 @@
...
+	l.sugar.Debug(vals...)
+}
+
+func (l *logger) Debugf(format string, vals ...interface{}) {
+	l.sugar.Debugf(format, vals...)
+}
+
+func (l *logger) Debugd(msg string, details ...interface{}) {
+	if len(details) == 1 {
+		l.logger.Debug(msg, zap.Any(detailsKey, details[0]))
```
kpango, line 149: do we need to unwrap the slice type if member equals 1.

internal/net/grpc/interceptor/server/logging/accesslog.go
```diff
@@ -0,0 +1,152 @@
...
+
+		service, method := parseMethod(info.FullMethod)
+
+		entity := &AccessLogEntity{
+			GRPC: &AccessLogGRPCEntity{
+				Kind:    grpcKindUnary,
+				Service: service,
+				Method:  method,
+			},
+			StartTime: startTime,
```
kpango, line 80: I think there are cases where you want to record the end time like the CreateIndex phase.
You can calculate end time from startTime and latency, but since time.Now is called internally in time.Since, you should record the end time with time.Now and then calculate Latency using time.Sub. How about changing it?

internal/runner/runner.go
```diff
@@ -71,18 +71,21 @@ func Do(ctx context.Context, opts ...Option) error {
 
 	if isHelp || err != nil {
 		log.Init(log.WithLevel(level.FATAL.String()))
+		defer log.Close()
```
kpango, line 74: I think log.Fatal in main.go may crash if you close log.Object here.

kpango (changes requested):
I never thought that you implement Infod like klog's InfoS, seems good idea.

internal/log/zap/zap.go
rinx: We should do it here because the given `details` is treated as slice and then zap will log it as the format like:
```
{"level": "debug", ... , "details": [{"foo": "bar", ...}]}
```
we want here is:
```
{"level": "debug", ... , "details": {"foo": "bar", ...}}
```

rinx: Or should we revise the interface of `debugd` like below?
```
Debugd(msg string, details interface{})
```

kpango: hmm, I think it's okay golang can't use override.
I understand about background, it's okay.

internal/log/glg/glg.go
rinx: done.

internal/net/grpc/interceptor/server/logging/accesslog.go
rinx: I've just added `endTime` field. please check it.

internal/runner/runner.go
rinx: I removed them. Is it okay?

kpango: I think it's okay but we have to close logger, so I think it's not beautiful answer but we should close it after main.go's log.Fatal

internal/net/grpc/interceptor/server/logging/accesslog.go
kpango: checked ! thanks!.

internal/runner/runner.go
rinx: ah, i got it.

kpango: oh, I gotta beautiful answer,,,,

kpango: we should close it in the log.Fatal Function.

rinx: That's right.

kpango: log.Fatal kills process we can close it

rinx: But there may be same thing here, if the runner.Do returns some error?
https://github.com/vdaas/vald/blob/fc1e535f5ed8ac4042fce532d5cbe7746b1fa470/internal/runner/runner.go#L99

kpango: omg, I make mistale, that's true,,,,

kpango: hmm, ok, I think current implementation is correct, beacuse if log.Fatal runs we cannot use defer.
I'll approve this PR.

kpango:
can you please check CI (test/internal & golangci-lint)

internal/runner/runner.go
rinx: i pointed that if runner.Do returns some error, `defer log.Close()` at L99 runner.go will be executed before `log.Fatal` in the cmd/xxx/main.go.
I think the `log.Fatal` may crash.

kpango: I think it depends on the log library.
glg Close func doesn't do anything, so it doesn't affect log.Fatal panic.
and zap only Syncs to flush the log buffer, so I don't think there is a problem with zap.

rinx: I see.

rinx:
resolved.

kpango (approved):
LGTM

PR #945, opened 2021-01-21
Add option to use networking.k8s.io/v1beta1 ingresses

rinx:
### Description:

- Add `usev1beta1` option to use networking.k8s.io/v1beta1 version of ingresses.
- Remove useless values yaml
- Fix agent-standalone yaml

### Related Issue:

Nothing

Nothing

### Environment:

- Go Version: 1.15.7
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.13.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [X] Refactoring [type/refactoring]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


charts/vald/values.yaml
```diff
@@ -995,6 +995,9 @@ gateway:
       # @schema {"name": "gateway.vald.ingress.servicePort", "type": "string"}
       # gateway.vald.ingress.servicePort -- service port to be exposed by ingress
       servicePort: grpc
+      # @schema {"name": "gateway.vald.ingress.usev1beta1", "type": "boolean"}
+      # gateway.vald.ingress.usev1beta1 -- use networking.k8s.io/v1beta1 instead of v1
+      usev1beta1: false
```
rinx, line 1000: Currently the default value is `false`. what do you think about it?

kevindiu:
I am thinking the same issue may happen in later k8s also, maybe we can let user to choose which version to use in other k8s resources also.

charts/vald/values.yaml
kpango: I think it's okay, and we should add comment if k8s 1.22 released it would be better to remove this option

charts/vald/templates/gateway/backup/ing.yaml
```diff
@@ -15,7 +15,11 @@
 #
 {{- $gateway := .Values.gateway.backup -}}
 {{- if and $gateway.enabled $gateway.ingress.enabled }}
+{{- if $gateway.ingress.usev1beta1 }}
+apiVersion: networking.k8s.io/v1beta1
```
kevindiu, line 19: which one do you think is better?
```suggestion
apiVersion: {{ $gateway.ingress.apiVersion }}
```
and use set the string `networking.k8s.io/v1beta1` to the values.yaml

kpango:
how about define vald wide api version definition list in the charts's defaults section like
```
defaults:
    k8s:
        apiVersion:
            k8s: v1
            apps: apps/v1
            apiextensions-k8s-io: apiextensions.k8s.io/v1
            batch: batch/v1
            policy: policy/v1beta1
            scheduling-k8s-io: scheduling.k8s.io/v1
            autoscaling: v2beta2
            networking-k8s-io: networking.k8s.io/v1
            rbac-authorization-k8s-io: rbac.authorization.k8s.io/v1
```
and use it
```
{{- $defaults := .Values.defaults -}}
{{- $gateway := .Values.gateway.lb -}}
{{- if and $gateway.enabled $gateway.ingress.enabled }}
apiVersion: {{ $defaults.k8s.apiVersion.networking-k8s-io }}
kind: Ingress
```

kevindiu:
> if some api version breaks backward compatibility we have to add logical section.
I starting thinking we should manage k8s manifest by version

kpango:
> I starting thinking we should manage k8s manifest by version

yes we should care about it, cuz we provide helm charts for user usually charts provider maintains k8s apiVersion.

rinx:
I think it's not a good idea to allow users to edit api-versions through values.yaml.
It takes too much cost to maintain. Keeping backward compatibility would be really hard.

If the users want to edit them, they should use `helm template` command to render manifests and edit them manually.

rinx:
Maybe we should define required version of K8s for each version of the charts.

kpango:
I think that specifying apiVersion from values.yaml and implementing conditional branching with usev1beta1 are not so different in maintenance cost, is there any other cost that I am not considering?

kpango:
But, as you mentioned, we shouldn't support backward compatibility about api version. in this document which said helm automatically resolves api version which suitable for current k8s context cluster version.
https://helm.sh/docs/topics/kubernetes_apis/

rinx:
> usev1beta1 are not so different in maintenance cost

that's right.

I thought this PR is just for temporary workaround for the problem that is happened in our dev cluster.
After resolved the problem, I'd like to remove the conditional branches from our chart.

kpango:
> After resolved the problem, I'd like to remove the conditional branches from our chart.

I understand that this pull request will be used as a temporary workaround.
Also, I think there is no use case for apiVersion to be different for each component, I don't think it is necessary to have a setting for each component, but only for defaults.

kevindiu:
How about this approach?
we will have different version of yaml stored and user can use different version of it. 
e.g. we store multiple version of `ing.yaml` such as `ing-v1.yaml` and manage them one by one?

kpango:
@kevindiu we can't do that, helm operator will apply everything that he made.

rinx:
In the private chat, Kpango said that it is enough to prepare `usev1beta1` only in the defaults section. I agree with him.
So I'm going to revise this PR to have a `usev1beta1` in the defaults section and add some comments on that it'll be removed after becoming useless.

rinx:
I've just updated. Could you please check the changes? @kevindiu @kpango

kpango:
okay, and I opened new issue #946 about this pr's problems.

kpango (approved):
LGTM

PR #996, opened 2021-02-09
fix MySQL panic test

vankichi:
Fix the panic error of MySQL unit test.

### Description:
I researched the panic error during the MySQL unit test.
This error may be caused by `nil` pointer of the session which is one of the MySQL client struct values.
My  change is following:
- added the nil check for each function that uses `session`.
- added new error type for MySQL: `ErrMySQLSessionNil`
- fix the code for using session mock.

I need the authors' review.

### Environment:

- Go Version: 1.15.7
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.13.1

### Types of changes:

- [x] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


internal/errors/mysql.go
```diff
@@ -50,6 +50,11 @@ var (
 	ErrRequiredMemberNotFilled = func(member string) error {
 		return Wrapf(NewErrMySQLInvalidArgumentIdentity(), "error required member not filled (member: %s)", member)
 	}
+
+	// ErrMySQLSessionNil represents a function to generate an error that the MySQL session is nil.
+	ErrMySQLSessionNil = func() error {
```
kpango, line 55: I think this case you shouldn't use functional type for error, but if you add additional information such as below it's better
```
	db                   string
	host                 string
	port                 int
	user                 string
	pass                 string
	name                 string
	charset              string
```

vankichi: @kpango so, for example, should I implement the below code instead of this?
```go
if m.session == nil {
  return errors.Errorf("session is nil, { db: %s, host: %s, port: %d, user: %s, pass: %s, name: %s, charset: %s", m.db, m.host, m.port, m.user, m.pass, m.name, m.charset)
}
```

kpango:
```suggestion
	ErrMySQLSessionNil = func(db, host, user, name, charset string, port int) error {
		return errors.Errorf("session is nil, { db: %s, host: %s, port: %d, user: %s, name: %s, charset: %s", db, host, port, user, name, charset)
```

kpango: or you can print them on debug log.

kpango: I think debug log is easy.

vankichi: @kpango thx for the review, I updated the code 🙏

internal/db/rdb/mysql/mysql.go
```diff
@@ -168,8 +178,17 @@ func (m *mySQLClient) Ping(ctx context.Context) (err error) {
...
 func (m *mySQLClient) Close(ctx context.Context) (err error) {
+	if m.session == nil {
+		err = errors.ErrMySQLSessionNil
+		log.Debugf(
+			"err: %s, { db: %s, host: %s, port: %d, user: %s, pass: %s, name: %s, charset: %s } ",
+			err.Error(), m.db, m.host, m.port, m.user, m.pass, m.name, m.charset,
+		)
+		return err
+	}
+
```
kevindiu, line 191: I think we should not return error when session is nil, do nothing instead.
```suggestion
```

internal/db/rdb/mysql/mysql.go
```diff
@@ -168,8 +178,17 @@ func (m *mySQLClient) Ping(ctx context.Context) (err error) {
...
 	if m.connected.Load().(bool) {
```
kevindiu, line 192:
```suggestion
	if m.session != nil && m.connected.Load().(bool) {
```

kevindiu (changes requested):

kevindiu:
missing session validation in `SetVector()`.
https://github.com/vdaas/vald/pull/996/files#diff-3b634a30e03f37629a7475da00b816954f658f6ee80a63c56f74a93de34a2795R288

kevindiu:
please also add `author review required` to the PR description :)

internal/db/rdb/mysql/mysql.go
```diff
@@ -141,6 +143,14 @@ func (m *mySQLClient) Open(ctx context.Context) (err error) {
 // Ping check the connection of MySQL database.
 // If the connection is closed, it returns error.
 func (m *mySQLClient) Ping(ctx context.Context) (err error) {
+	if m.session == nil {
+		err = errors.ErrMySQLSessionNil
+		log.Debugf(
+			"err: %s, { db: %s, host: %s, port: %d, user: %s, pass: %s, name: %s, charset: %s } ",
+			err.Error(), m.db, m.host, m.port, m.user, m.pass, m.name, m.charset,
```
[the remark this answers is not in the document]
kevindiu: we should not print the password to the log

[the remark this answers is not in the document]
vankichi: @kevindiu 
why did you suggest as above?

internal/db/rdb/mysql/mysql.go
vankichi: @kevindiu why not?
if the session is nil, it would be occurred panic

internal/db/rdb/mysql/mysql.go
vankichi: 🤔

internal/db/rdb/mysql/mysql.go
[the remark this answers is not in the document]
kevindiu: I think we can use `err` instead of `err.Error()` in `Debugf()`.
https://stackoverflow.com/questions/43887776/what-is-the-correct-format-specifier-to-print-error-object-in-go-s-or-v

internal/db/rdb/mysql/mysql.go
kevindiu: but it is `Close()` function. If the session is nil I think it is better to return.
What do you think?

internal/db/rdb/mysql/mysql.go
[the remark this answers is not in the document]
vankichi: it is correct, but we use glg as a log package.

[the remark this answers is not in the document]
hlts2: The final output is the same, so I think it doesn't matter which one 🤔

internal/db/rdb/mysql/mysql.go
```diff
@@ -168,8 +178,17 @@ func (m *mySQLClient) Ping(ctx context.Context) (err error) {
...
 // Close closes the connection of MySQL database.
-// If the connection is already closed or closing conncection is failed, it returns error.
+// If the connection is already closed or closing connection is failed, it returns error.
 func (m *mySQLClient) Close(ctx context.Context) (err error) {
+	if m.session == nil {
+		err = errors.ErrMySQLSessionNil
+		log.Errorf(
+			"err: %s, { db: %s, host: %s, port: %d, user: %s, name: %s, charset: %s } ",
+			err.Error(), m.db, m.host, m.port, m.user, m.name, m.charset,
+		)
```
hlts2, line 188: IMO:
Since the same code appears multiple times, I thought it would be cleaner if I created a new function and formatted it and output an error inside it.
if I made a mistake, I am sorry.

internal/db/rdb/mysql/mysql.go
kevindiu: @kpango @vankichi what do you think?

internal/db/rdb/mysql/mysql.go
vankichi: @hlts2 e. g. below code?
```go
func (m *mySQLClient) outputNilSessionLog(err error) {
	log.Errorf(
		"err: %s, { db: %s, host: %s, port: %d, user: %s, name: %s, charset: %s } ",
		err.Error(), m.db, m.host, m.port, m.user, m.name, m.charset,
	)}
}
```

hlts2: yes, It looks good to me!

internal/db/rdb/mysql/mysql.go
```diff
@@ -172,8 +182,17 @@ func (m *mySQLClient) Ping(ctx context.Context) (err error) {
...
```
kevindiu, line 192:
```suggestion
		)
		m.connected.Store(false)
```

kevindiu (approved):
LGTM

hlts2 (approved):
thank you :bow:
LGTM

kpango (approved):
LGTM

PR #1028, opened 2021-02-19
remove vcache for vald agent due to vcache delete timing control failure and time ordered concurrent vector queue called vqueue

kpango:

### Description:
We are currently experiencing a bug where the Vald agent is unable to ensure the integrity of vector Delete and Insert when performing high frequency Upsert and Delete in a multi-threaded environment.
The Vald Agent was using Vector Cache = vcache to reduce the overhead of calling FFI inline in user requests, and also to aggregate FFI calls during CreateIndex.

However, since vcache uses a sync.Map-like implementation and does not support ordered loops, depending on the timing, the old vcache may not be deleted and may be loaded during the next CreateIndex.
In addition, Insert Vector Cache (IVC) and Delete Vector Cache (DVC) compare the timestamps of execution plans with the same UUID and delete unnecessary caches to avoid duplication of FFIs.
I believe that this reciprocal cache deletion logic further expands the garbage data and the Delete and Insert operations were not executed.
To solve this problem, we need to removed the vcache and added a cache layer to ensure orderliness.

So, I implemented TOCVQ (Time Ordered Concurrent Vector Queue) to avoid this problem, always calculating the execution plan from the head of the queue (operating on the old time axis) and calling FFI calls to strongly guarantee the consistency of the results.

### Environment:

- Go Version: 1.16
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [x] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


pkg/agent/core/ngt/service/vqueue/queue.go
```diff
@@ -0,0 +1,326 @@
...
+		udm[d.uuid] = d.date
+	}
+	dl = dl[:0]
+	// we should check insert vqueue if insert vqueue exists and delete operation date is newer than insert operation date then we should remove insert vqueue's data.
+	v.imu.Lock()
+	for i, idx := range v.uii {
+		// check same uuid & operation date
+		// if date is equal, it may update operation we shouldn't remove at that time
+		date, exists := udm[idx.uuid]
+		if exists && date > idx.date {
```
rinx, line 297: I'm not sure but is it needed to remove `idx.uuid` entry from `udk` here?

pkg/agent/core/ngt/service/ngt.go
```diff
@@ -63,7 +64,6 @@ type NGT interface {
 	NumberOfCreateIndexExecution() uint64
 	NumberOfProactiveGCExecution() uint64
 	UUIDs(context.Context) (uuids []string)
-	UncommittedUUIDs() (uuids []string)
 	DeleteVCacheLen() uint64
 	InsertVCacheLen() uint64
```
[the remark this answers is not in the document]
kpango: Since this PR is already huge, can I just leave a TODO comment and add it to the project as a separate task for Core+SRE?

pkg/agent/core/ngt/service/vqueue/queue.go
```diff
@@ -0,0 +1,326 @@
...
+	RangePopDelete(ctx context.Context, f func(uuid string) bool)
+	IVQLen() int
+	DVQLen() int
+}
+
+type vqueue struct {
+	ich        chan index
+	uii        []index // un inserted index
+	imu        sync.Mutex
+	uiil       map[string][]float32
```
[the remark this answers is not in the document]
kpango: it means uninserted index list but I notice it's map structure so I changed uiil to uiim and add the comment!

and uiim is used for fetching vector data from vqueue.

pkg/agent/core/ngt/service/vqueue/queue.go
kpango: Good question.
In the CreateIndex operation of the NGT Service, the Delete Queue is processed first, and then the Insert Queue is processed, so the Insert Queue still contains the old Insert Operation older than the Delete Queue, and it is possible that data that was intended to be deleted is registered again.
For this reason, the data is deleted from the Insert Queue only when retrieving data from the Delete Queue. 
In the case of old implementation called VCache, the opposite process (delete Delete queue when retrieving Insert Queue)  was also performed, but this time it is not performed because there were many problems in update operations such as Update if Delete is deleted.

I'll add some comment about it.

pkg/agent/core/ngt/service/ngt.go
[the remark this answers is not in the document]
kpango: sorry I checked the point that I have to rename, looks not so difficult can I change it in this PR?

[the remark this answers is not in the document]
rinx: okay, but please add old promQLs to Grafana dashboards for keeping backward compatibility.

pkg/agent/core/ngt/service/vqueue/queue.go
rinx: I know about that, so I just wondered that `udk` will have an unexisting key after removing insert vqueue entry by `dl` index. I asked is it okay that `udk` is not updated here.

kpango: sorry I don't get it, do you mean should I remove like `delete(udk, idx.uuid)` ????

pkg/agent/core/ngt/service/ngt.go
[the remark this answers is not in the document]
kpango: you mean should I rename this line?
or 
add another target ?
https://github.com/vdaas/vald/blob/master/k8s/metrics/grafana/dashboards/01-vald-agent.yaml#L857-L863

pkg/agent/core/ngt/service/vqueue/queue.go
rinx: right.
Doesn't it make sense?

kpango: No, it doesn't. If you delete the udk entry here, it will cause deletion omissions, and the delete instruction should pass through all of them, as it did in the vcahce bug mentioned above.
If there is a newer Insert operation than the Delete instruction, the Delete instruction should also be executed.

pkg/agent/core/ngt/service/ngt.go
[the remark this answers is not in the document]
rinx: I've just noticed that they were used in Stats Panels of the Grafana visualization. It cannot treat multiple queries...
So it is okay to not support backward compatibility here.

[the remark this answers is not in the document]
kpango: thanks, I'll just rename them.

pkg/agent/core/ngt/service/vqueue/queue.go
rinx: I got. I've misunderstood because I forgot to think about L279-287 that removes duplicated entries in Delete Vqueue.

kpango: I see, sorry for really complex code,,, I'd like to refactor in the far future... :sweat_smile:

pkg/agent/core/ngt/handler/grpc/handler.go
```diff
@@ -1175,7 +1175,7 @@ func (s *server) IndexInfo(ctx context.Context, _ *payload.Empty) (res *payload.
 	}()
 	return &payload.Info_Index_Count{
 		Stored:      uint32(s.ngt.Len()),
-		Uncommitted: uint32(s.ngt.InsertVQueueLen()),
+		Uncommitted: uint32(s.ngt.InsertVQueueBufferLen()),
```
rinx, line 1178: Doesn't need to add `InsertVQueueChannelLen`?

kpango: In this case, `InsertVQueueChannelLen` data is not ready for create index, so we can ignore it at this time, but we should add `DeleteVQueueBufferLen` thank you for mention it.

rinx: I see.

rinx:
Seems okay to merge but please ask to the others.

kpango:
@vankichi @hlts2 @kevindiu @datelier can you please review this PR?

vankichi (dismissed):

pkg/agent/core/ngt/service/ngt.go
```diff
@@ -569,89 +542,46 @@ func (n *ngt) CreateIndex(ctx context.Context, poolSize uint32) (err error) {
...
-	for _, duuid := range delList {
-		n.dvc.Delete(duuid)
-		id, ok := n.kvs.Delete(duuid)
-		if !ok {
-			log.Error(errors.ErrObjectIDNotFound(duuid).Error())
-			err = errors.Wrap(err, errors.ErrObjectIDNotFound(duuid).Error())
+	log.Info("create index delete phase finished")
+	n.gc()
+	log.Info("create index insert phase started")
+	n.vq.RangePopInsert(ctx, func(uuid string, vector []float32) bool {
```
kevindiu, line 574: I am not sure if it is correct:
The current logic is process all the delete request first, and insert request later

For example we have 2 request, reqA and reqB
reqA: insert vecA request, time t+0
reqB: delete vecA request, time t+10

If they are processing the same vector, the final result will be the vecA is inserted, but maybe it is not correct.

Am I correct?

hlts2 (approved):
LGTM

pkg/agent/core/ngt/service/ngt.go
kpango: that's why I remove the Insert Queue here.
https://github.com/vdaas/vald/pull/1028/files#diff-d8459a99ea4f8ec96e01405d5f55171070363986154ad61e269c7911dbfd91a9R304-R327

kevindiu: I see... I think it is good.

kevindiu (dismissed):
LGTM

PR #1086, opened 2021-03-16
Add benchmark for internal/timeutil/time.go

hlts2:

### Description:

I added a benchmark for `timeutil/time.go` with table-driven.

```
goos: linux
goarch: amd64
pkg: github.com/vdaas/vald/internal/timeutil
cpu: Intel(R) Core(TM) i7-6920HQ CPU @ 2.90GHz
BenchmarkParse/t_1ns-7        	123051638	         9.610 ns/op	       0 B/op	       0 allocs/op
BenchmarkParse/t_1µs-7        	100000000	        10.14 ns/op	       0 B/op	       0 allocs/op
BenchmarkParse/t_1ms-7        	94272541	        12.00 ns/op	       0 B/op	       0 allocs/op
BenchmarkParse/t_1s-7         	100000000	        10.12 ns/op	       0 B/op	       0 allocs/op
BenchmarkParse/t_1m-7         	100000000	        11.35 ns/op	       0 B/op	       0 allocs/op
BenchmarkParse/t_1h-7         	95456136	        12.64 ns/op	       0 B/op	       0 allocs/op
BenchmarkParseWithDefault/t_1ns_and_d_50_millisecond-7         	144522436	         8.491 ns/op	       0 B/op	       0 allocs/op
BenchmarkParseWithDefault/t_1µs_and_d_50_millisecond-7         	143124103	         8.882 ns/op	       0 B/op	       0 allocs/op
BenchmarkParseWithDefault/t_1ms_and_d_50_millisecond-7         	100000000	        10.55 ns/op	       0 B/op	       0 allocs/op
BenchmarkParseWithDefault/t_1s_and_d_50_millisecond-7          	137396325	         9.680 ns/op	       0 B/op	       0 allocs/op
BenchmarkParseWithDefault/t_1m_and_d_50_millisecond-7          	117594116	         9.829 ns/op	       0 B/op	       0 allocs/op
BenchmarkParseWithDefault/t_1h_and_d_50_millisecond-7          	99942590	        10.91 ns/op	       0 B/op	       0 allocs/op
BenchmarkParseWithDefault/t_1days_and_d_50_millisecond-7       	 3074391	       392.0 ns/op	     336 B/op	       9 allocs/op
BenchmarkParseWithDefault/t_1month_and_d_50_millisecond-7      	 2906296	       410.0 ns/op	     368 B/op	       9 allocs/op
BenchmarkParseWithDefault/t_10month_and_d_50_millisecond-7     	 2822518	       442.3 ns/op	     368 B/op	       9 allocs/op
BenchmarkParseWithDefault/t_100month_and_d_50_millisecond-7    	 2812122	       427.2 ns/op	     384 B/op	       9 allocs/op
BenchmarkParseWithDefault/t_1000month_and_d_50_millisecond-7   	 2759754	       426.8 ns/op	     384 B/op	       9 allocs/op
BenchmarkParseWithDefault/t_10000month_and_d_50_millisecond-7  	 2839292	       460.2 ns/op	     384 B/op	       9 allocs/op
PASS
ok  	github.com/vdaas/vald/internal/timeutil	29.154s
```

### Environment:

- Go Version: 1.16
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/timeutil/time_bench_test.go
+++ b/internal/timeutil/time_bench_test.go
@@ -1,0 +1,172 @@
+package timeutil
+
+import (
+	"reflect"
+	"testing"
+	"time"
+)
+
+func BenchmarkParse(b *testing.B) {
+	type args struct {
+		t string
+	}
+	type want struct {
+		want time.Duration
+		err  error
+	}
+	type test struct {
+		name string
+		args args
+		want want
+	}
+
+	tests := []test{
+		{
+			name: "when t is 10ms",
+			args: args{
+				t: "10ms",
+			},
+			want: want{
+				want: 10 * time.Millisecond,
+				err:  nil,
+			},
+		},
+		{
+			name: "when t is 100ms",
+			args: args{
+				t: "100ms",
+			},
+			want: want{
+				want: 100 * time.Millisecond,
+				err:  nil,
+			},
+		},
+		{
+			name: "when t is 1s",
+			args: args{
+				t: "1s",
+			},
+			want: want{
+				want: time.Second,
+				err:  nil,
+			},
+		},
+		{
+			name: "when t is 10s",
+			args: args{
+				t: "10s",
+			},
+			want: want{
+				want: 10 * time.Second,
+				err:  nil,
+			},
+		},
+		{
+			name: "when t is 100s",
+			args: args{
+				t: "100s",
+			},
+			want: want{
+				want: 100 * time.Second,
+				err:  nil,
+			},
+		},
+	}
+
+	for _, test := range tests {
+		b.Run(test.name, func(b *testing.B) {
+			b.RunParallel(func(pb *testing.PB) {
+				for pb.Next() {
+					got, err := Parse(test.args.t)
+					if err != nil {
+						b.Error(err)
+					}
+					if !reflect.DeepEqual(test.want.want, got) {
+						b.Errorf("want: %v, but got: %v", test.want.want, got)
+					}
+				}
+			})
+		})
+	}
+}
+
+func BenchmarkParseWithDefault(b *testing.B) {
+	type args struct {
+		t string
+		d time.Duration
+	}
+	type want struct {
+		want time.Duration
+	}
+	type test struct {
+		name string
+		args args
+		want want
+	}
+
+	tests := []test{
+		{
+			name: "when t is 10second",
+			args: args{
+				t: "10second",
+				d: 50 * time.Millisecond,
+			},
+			want: want{
+				want: 50 * time.Millisecond,
+			},
+		},
+		{
+			name: "when t is 100second",
+			args: args{
+				t: "100second",
+				d: 50 * time.Millisecond,
+			},
+			want: want{
+				want: 50 * time.Millisecond,
+			},
+		},
+		{
+			name: "when t is 1000second",
+			args: args{
+				t: "1000second",
+				d: 50 * time.Millisecond,
+			},
+			want: want{
+				want: 50 * time.Millisecond,
+			},
+		},
+		{
+			name: "when t is 10000second",
+			args: args{
+				t: "1000second",
+				d: 50 * time.Millisecond,
+			},
+			want: want{
+				want: 50 * time.Millisecond,
+			},
+		},
+		{
+			name: "when t is 100000second",
+			args: args{
+				t: "10000second",
+				d: 50 * time.Millisecond,
+			},
+			want: want{
+				want: 50 * time.Millisecond,
+			},
+		},
+	}
+
+	for _, test := range tests {
+		b.Run(test.name, func(b *testing.B) {
+			b.RunParallel(func(pb *testing.PB) {
+				for pb.Next() {
+					got := ParseWithDefault(test.args.t, test.args.d)
+					if !reflect.DeepEqual(test.want.want, got) {
+						b.Errorf("want: %v, but got: %v", test.want.want, got)
+					}
+				}
+			})
+		})
+	}
+}
--- a/internal/timeutil/time_bench_test.go
+++ b/internal/timeutil/time_bench_test.go
@@ -1,7 +1,6 @@
 package timeutil
 
 import (
-	"reflect"
 	"testing"
 	"time"
 )
@@ -10,14 +9,9 @@ func BenchmarkParse(b *testing.B) {
 	type args struct {
 		t string
 	}
-	type want struct {
-		want time.Duration
-		err  error
-	}
 	type test struct {
 		name string
 		args args
-		want want
 	}
 
 	tests := []test{
@@ -26,19 +20,11 @@ func BenchmarkParse(b *testing.B) {
 			args: args{
 				t: "10ms",
 			},
-			want: want{
-				want: 10 * time.Millisecond,
-				err:  nil,
-			},
 		},
 		{
 			name: "when t is 100ms",
 			args: args{
 				t: "100ms",
-			},
-			want: want{
-				want: 100 * time.Millisecond,
-				err:  nil,
 			},
 		},
 		{
@@ -46,29 +32,17 @@ func BenchmarkParse(b *testing.B) {
 			args: args{
 				t: "1s",
 			},
-			want: want{
-				want: time.Second,
-				err:  nil,
-			},
 		},
 		{
 			name: "when t is 10s",
 			args: args{
 				t: "10s",
 			},
-			want: want{
-				want: 10 * time.Second,
-				err:  nil,
-			},
 		},
 		{
 			name: "when t is 100s",
 			args: args{
 				t: "100s",
-			},
-			want: want{
-				want: 100 * time.Second,
-				err:  nil,
 			},
 		},
 	}
@@ -80,9 +54,10 @@ func BenchmarkParse(b *testing.B) {
 					got, err := Parse(test.args.t)
 					if err != nil {
 						b.Error(err)
+						return
 					}
-					if !reflect.DeepEqual(test.want.want, got) {
-						b.Errorf("want: %v, but got: %v", test.want.want, got)
+					if got == 0 {
+						b.Error("got is 0")
 					}
 				}
 			})
@@ -95,13 +70,9 @@ func BenchmarkParseWithDefault(b *testing.B) {
 		t string
 		d time.Duration
 	}
-	type want struct {
-		want time.Duration
-	}
 	type test struct {
 		name string
 		args args
-		want want
 	}
 
 	tests := []test{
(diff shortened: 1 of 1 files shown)


vankichi:
we discussed do not implement bench code like table-driven tests before.
your implement seems like as table-driven tests.
could you fix the right coding that we discussed?

vankichi:
@hlts2 cc: @kevindiu 
> **I wanted to take a bench mark with multiple parameters.
When I was creating a top-level bench function for each parameter, the number of benchmark function increased.**

I know what you'd like to do.
And this problem is obvious when we decided we do not implement like table-drive tests.
Table-driven bench code cannot cover all patterns
Also, it will be complex when we use both table-driven test style and another style for bench code.
I think it is not good for the maintenance, and more the bench code will be depended on the contributor.
It is the reason I don't suppose this way and we should implement the most simple style.

After reading this, what do you think?
And if you'd like to implement like as table-driven test, please explain more about the merit and how we handle the bench coding which we cannot use table-driven style.

hlts2:
@vankichi 
> we discussed do not implement bench code like table-driven tests before.
your implement seems like as table-driven tests.
could you fix the right coding that we discussed?

sorry. I will fix.
But Is table-driven really suitable in this case?
For example, If we want to benchmark 10 functions, and when they have one parameter, do we create as many top-level benchmark functions as there are parameters?
If you want each parameter to change 3 values, you need to write a total of 30 top-level benchmark functions like this. Is this correct

```
// implementation code
func A(s string) {}
func B(s string) {}
func C(s string) {}

// if we want to create benchmark for above function, we will create the following function ....
// benchmark code
func BencharkA_s_is_hoge() {}
func BencharkA_s_is_foo() {}
func BencharkA_s_is_bar() {}

func BencharkB_s_is_hoge() {}
func BencharkB_s_is_foo() {}
func BencharkB_s_is_bar() {}

...

```

vankichi:
@hlts2 @kevindiu 
Actually, it is suitable in this case.
But, at once. could you read this comment carefully and think more deeply?
https://github.com/vdaas/vald/pull/1086#issuecomment-800752871

> For example, If we want to benchmark 10 functions, and when they have one parameter, do we create as many top-level benchmark functions as there are parameters?
If you want each parameter to change 3 values, you need to write a total of 30 top-level benchmark functions like this. Is this correct?

I think it seems to the edge case.
I cannot say this situation will not occur, but before creating a bench code, we have to care about the purpose of the benchmark whether we should test the all of patterns that you mentioned.
If we should, I think we should create as code that you showed us.

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,172 @@
...
+		b.Run(test.name, func(b *testing.B) {
+			b.RunParallel(func(pb *testing.PB) {
+				for pb.Next() {
+					got, err := Parse(test.args.t)
+					if err != nil {
+						b.Error(err)
+					}
+					if !reflect.DeepEqual(test.want.want, got) {
+						b.Errorf("want: %v, but got: %v", test.want.want, got)
+					}
```
kevindiu, line 86: I think we shouldn't check the result of the return value. The reasons are:
- we have already tested in unit test and I think it  is better to separate the concern of unit test and benchmark
- this may affect the result of benchmarking as it adds additional cost to execute the benchmark

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+func BenchmarkParse(b *testing.B) {
+	type args struct {
+		t string
+	}
+	type test struct {
+		name string
+		args args
+	}
+
```
kevindiu, line 16:
```suggestion
```

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,128 @@
...
+			},
+		},
+		{
+			name: "t 100s",
+			args: args{
+				t: "100s",
+			},
+		},
+	}
+
```
kevindiu, line 49:
```suggestion
```

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,128 @@
...
+func BenchmarkParseWithDefault(b *testing.B) {
+	type args struct {
+		t string
+		d time.Duration
+	}
+	type test struct {
+		name string
+		args args
+	}
+
```
kevindiu, line 77:
```suggestion
```

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+		{
+			name: "t 100000second and 50*time.Millisecond",
+			args: args{
+				t: "10000second",
+				d: 50 * time.Millisecond,
+			},
+		},
+	}
+
```
kevindiu, line 115:
```suggestion
```

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,128 @@
...
+		d time.Duration
+	}
+	type test struct {
+		name string
+		args args
+	}
+
+	tests := []test{
+		{
+			name: "t 10second and 50*time.Millisecond",
```
kevindiu, line 80:
```suggestion
			name: "t 10 second and d 50 millisecond",
```

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,128 @@
...
+			args: args{
+				t: "10second",
+				d: 50 * time.Millisecond,
+			},
+		},
+		{
+			name: "t 100second and 50*time.Millisecond",
```
kevindiu, line 87:
```suggestion
			name: "t 100 second and d 50 millisecond",
```

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,128 @@
...
+			args: args{
+				t: "100second",
+				d: 50 * time.Millisecond,
+			},
+		},
+		{
+			name: "t 1000second and 50*time.Millisecond",
```
kevindiu, line 94: same

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,128 @@
...
+			args: args{
+				t: "1000second",
+				d: 50 * time.Millisecond,
+			},
+		},
+		{
+			name: "t 10000second and 50*time.Millisecond",
```
kevindiu, line 101: same

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,128 @@
...
+			args: args{
+				t: "1000second",
+				d: 50 * time.Millisecond,
+			},
+		},
+		{
+			name: "t 100000second and 50*time.Millisecond",
```
kevindiu, line 108: same

internal/timeutil/time_bench_test.go
hlts2, line 94:
```suggestion
			name: "t 1000 second and 50 time.Millisecond",
```

internal/timeutil/time_bench_test.go
hlts2, line 101:
```suggestion
			name: "t 10000 second and d 50 time.Millisecond",
```

internal/timeutil/time_bench_test.go
hlts2, line 108:
```suggestion
			name: "t 100000 second and d 50 Millisecond",
```

internal/timeutil/time_bench_test.go
hlts2: thank you for your comment.
I fixed.

vankichi:
@hlts2 plz update the result of benchmark

hlts2:
@vankichi 
thank you. I updated 🙇

vankichi:
@hlts2 
how about adding more patterns?
e.g. `ns`, `h`, `days`, `month`

hlts2:
@vankichi 
> how about adding more patterns?
e.g. ns, m, h, days, month

I agree with you.
but an error occurs when the parsing `days` and `month`.
so I will add test pattern for `ns`, `m` `h`.
what do you think?

hlts2:
@vankichi 
thank you.
I added: https://github.com/vdaas/vald/pull/1086/commits/98b001600b444b7130d83a51b9b7467623449f50

vankichi (approved):
LGTM

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+		{
+			name: "t 1h and d 50 millisecond",
+			args: args{
+				t: "1h",
+				d: 50 * time.Millisecond,
+			},
+		},
+
+		// The following are error pattern.
```
kevindiu, line 125: Why do we need error pattern here?

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,189 @@
...
+			args: args{
+				t: "10month",
+				d: 50 * time.Millisecond,
+			},
+		},
+		// The length of t is 8.
+		{
+			name: "t 100month and d 50 millisecond",
+			args: args{
+				t: "10month",
```
kevindiu, line 156:
```suggestion
				t: "100month",
```

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,189 @@
...
+				d: 50 * time.Millisecond,
+			},
+		},
+		// The length of t is 9.
+		{
+			name: "t 1000month and d 50 millisecond",
+			args: args{
+				t: "10month",
```
kevindiu, line 164:
```suggestion
				t: "1000month",
```

internal/timeutil/time_bench_test.go
```diff
@@ -0,0 +1,189 @@
...
+				d: 50 * time.Millisecond,
+			},
+		},
+		// The length of t is 10.
+		{
+			name: "t 10000month and d 50 millisecond",
+			args: args{
+				t: "10month",
```
kevindiu, line 172:
```suggestion
				t: "10000month",
```

internal/timeutil/time_bench_test.go
hlts2: If an error occurs, the default value will be set, so take that into consideration.

hlts2: btw
I've added the following error pattern cases to take into account the length of the string when an error occurs.

hlts2:
@kevindiu @vankichi 
I updated benchmark result :bow:

vankichi (approved):

kevindiu (dismissed):
LGTM

PR #1091, opened 2021-03-17
Implement internal/timeutil/location/loc.go benchmark test

kevindiu:
Signed-off-by: kevindiu <kevin_diu@yahoo.com.hk>

This PR implement `internal/timeutil/location/loc.go` benchmark.

The benchmark result is:
```
goos: darwin
goarch: amd64
pkg: github.com/vdaas/vald/internal/timeutil/location
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkGMT-12    	  126840	      8916 ns/op	     352 B/op	       8 allocs/op
BenchmarkUTC-12    	1000000000	         0.9751 ns/op	       0 B/op	       0 allocs/op
BenchmarkJST-12    	   37604	     31617 ns/op	   57800 B/op	      14 allocs/op
PASS
ok  	github.com/vdaas/vald/internal/timeutil/location	3.988s
```

### Environment:

- Go Version: 1.16
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/timeutil/location/loc_bench_test.go
+++ b/internal/timeutil/location/loc_bench_test.go
@@ -1,3 +1,18 @@
+//
+// Copyright (C) 2019-2021 vdaas.org vald team <vald@vdaas.org>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
 package location
 
 import (
--- a/internal/timeutil/location/loc_bench_test.go
+++ b/internal/timeutil/location/loc_bench_test.go
@@ -1,0 +1,74 @@
+package location
+
+import (
+	"testing"
+	"time"
+)
+
+var loc *time.Location
+
+func BenchmarkGMT(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			GMT()
+		}
+	})
+}
+
+func BenchmarkGMT1(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	var l *time.Location
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			l = GMT()
+		}
+	})
+	loc = l
+}
+
+func BenchmarkUTC(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			UTC()
+		}
+	})
+}
+
+func BenchmarkUTC1(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	var l *time.Location
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			l = UTC()
+		}
+	})
+	loc = l
+}
+
+func BenchmarkJST(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			JST()
+		}
+	})
+}
+
+func BenchmarkJST1(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	var l *time.Location
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			l = JST()
+		}
+	})
+	loc = l
+}
--- a/internal/timeutil/location/loc_bench_test.go
+++ b/internal/timeutil/location/loc_bench_test.go
@@ -2,31 +2,18 @@ package location
 
 import (
 	"testing"
-	"time"
 )
-
-var loc *time.Location
 
 func BenchmarkGMT(b *testing.B) {
 	b.ReportAllocs()
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
-			GMT()
+			if l := GMT(); l == nil {
+				b.Error("GMT return nil")
+			}
 		}
 	})
-}
-
-func BenchmarkGMT1(b *testing.B) {
-	b.ReportAllocs()
-	b.ResetTimer()
-	var l *time.Location
-	b.RunParallel(func(pb *testing.PB) {
-		for pb.Next() {
-			l = GMT()
-		}
-	})
-	loc = l
 }
 
 func BenchmarkUTC(b *testing.B) {
@@ -34,21 +21,11 @@ func BenchmarkUTC(b *testing.B) {
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
 		for pb.Next() {
-			UTC()
+			if l := UTC(); l == nil {
+				b.Error("UTC return nil")
+			}
 		}
 	})
-}
-
-func BenchmarkUTC1(b *testing.B) {
-	b.ReportAllocs()
-	b.ResetTimer()
-	var l *time.Location
-	b.RunParallel(func(pb *testing.PB) {
-		for pb.Next() {
-			l = UTC()
-		}
-	})
-	loc = l
 }
 
 func BenchmarkJST(b *testing.B) {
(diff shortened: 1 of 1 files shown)


internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+	})
+}
+
+func BenchmarkGMT1(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	var l *time.Location
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			l = GMT()
```
hlts2, line 26:
```suggestion
			if l = GMT(); err == nil {
				b.Error("nil object")
			}
```

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+var loc *time.Location
+
+func BenchmarkGMT(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			GMT()
```
hlts2, line 15: Should we check if the returned object is nil or not? 🤔

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+	})
+	loc = l
+}
+
+func BenchmarkUTC(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			UTC()
```
hlts2, line 37: same
https://github.com/vdaas/vald/pull/1091/files#r596503147

internal/timeutil/location/loc_bench_test.go
kevindiu: I think we shouldn't check the result as it should be well tested in unit test.
I think it is not the responsibility of benchmark test.
What do you think?

hlts2: I got it. But How do you know if the benchmark is working properly? 🤔 
we don't need to check the details of the object, but I thought it would be okay to check the return values ​​nil or not nil.

hlts2: And I thought that if I didn't store it in a variable, the compiler would be optimized and the result would be different from the actual use case. 🤔 
If I made a mistake, I am sorry.

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"testing"
+	"time"
+)
+
+var loc *time.Location
+
+func BenchmarkGMT(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
```
vankichi, line 12:
```suggestion
```

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+		for pb.Next() {
+			GMT()
+		}
+	})
+}
+
+func BenchmarkGMT1(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	var l *time.Location
```
vankichi, line 23:
```suggestion
	var l *time.Location
	b.ReportAllocs()
	b.ResetTimer()
```

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			l = GMT()
+		}
+	})
+	loc = l
```
vankichi, line 29:
```suggestion
```

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
+package location
+
+import (
+	"testing"
+	"time"
+)
+
+var loc *time.Location
```
vankichi, line 8:
```suggestion
```

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+		for pb.Next() {
+			l = GMT()
+		}
+	})
+	loc = l
+}
+
+func BenchmarkUTC(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
```
vankichi, line 34:
```suggestion
```

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+		for pb.Next() {
+			UTC()
+		}
+	})
+}
+
+func BenchmarkUTC1(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	var l *time.Location
```
vankichi, line 45:
```suggestion
	var l *time.Location
	b.ReportAllocs()
	b.ResetTimer()
```

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			l = UTC()
+		}
+	})
+	loc = l
```
vankichi, line 51:
```suggestion
```

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func BenchmarkJST(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
```
vankichi, line 56:
```suggestion
```

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+		for pb.Next() {
+			JST()
+		}
+	})
+}
+
+func BenchmarkJST1(b *testing.B) {
+	b.ReportAllocs()
+	b.ResetTimer()
+	var l *time.Location
```
vankichi, line 67:
```suggestion
	var l *time.Location
	b.ReportAllocs()
	b.ResetTimer()
```

internal/timeutil/location/loc_bench_test.go
```diff
@@ -0,0 +1,74 @@
...
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			l = JST()
+		}
+	})
+	loc = l
```
vankichi, line 73:
```suggestion
```

internal/timeutil/location/loc_bench_test.go
kevindiu: @vankichi what do you think about this from @hlts2?
`actual usecase stored it to variable.`

vankichi: it is ok.
I think we change only this line,
```suggestion
			l := GMT()
```

kevindiu: @vankichi if we change that way, the compile error will be returned due to unused variable.
we may consider change it to:
```suggestion
        var l *time.Location
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			l = GMT()
```
cc: @hlts2

vankichi: or this? 🤔 
```suggestion
			_ := GMT()
```

kevindiu: @vankichi it is same as calling GMT() directly... 🤔
so it is `_ = GMT()` not `:=` 😵

vankichi: I think it is good to define a variable before `b.RunPrarallel`.
It seems to store same address 🤔

kevindiu: yes it store to the same address.. 
if you think it is good, I will update like this.
```go
        var l *time.Location
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			l = GMT()
```
cc: @hlts2

kevindiu: @vankichi @hlts2 since there is compile error if we do not use `l` in the above code, I change the implementation like this. What do you think?
```
func BenchmarkGMT(b *testing.B) {
	var l *time.Location
	b.ReportAllocs()
	b.ResetTimer()
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			l = GMT()
		}
	})
	_ = l
}
```

hlts2: @kevindiu 
In the first place, Is this code correct? 🤔 
Is it okay to rewrite the same variable from multiple Goroutines of RunParallel?

kevindiu: I think the cost of writing the same variable is small and we can ignore it, we do not put any lock on it so it should be ok.
what do you think? @hlts2

vankichi: I just tied but the above code won't work well.
did you check it?

hlts2 (dismissed):
it looks good to me!

vankichi (dismissed):
LGTM

PR #1117, opened 2021-03-25
create internal/config/cassandra test

vankichi:

### Description:

I created the test for `internal/config/cassandra.go`.
During creating the test, I found the error is ignored when `net.NewDialer()` returns the error.
So, I refctored it like as `internal/config/mysql`.

### Environment:

- Go Version: 1.16
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [ ] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


internal/config/cassandra_test.go
```diff
@@ -85,115 +87,253 @@ func TestCassandra_Bind(t *testing.T) {
...
+						KVTable:                  "kv",
+						VKTable:                  "vk",
+						VectorBackupTable:        "backup_vector",
+					},
+				},
+			}
+		}(),
+		func() test {
+			key := "PASSWORD"
+			val := "cassandra_password"
```
hlts2, line 285: Do you need other keys as well?:thinking:

hlts2 (commented):
I commented one. 🙏  but I think overall is good 👍

internal/config/cassandra_test.go
vankichi: 🤔  
it is not a must, only depends what you think

hlts2 (approved):

internal/config/cassandra_test.go
```diff
@@ -296,128 +442,355 @@ func TestCassandra_Opts(t *testing.T) {
...
+					VectorBackupTable:        "backup_vector",
+				},
+				beforeFunc: func(t *testing.T) {
+					t.Helper()
+					path := os.ExpandEnv("$GOROOT") + "/src/crypto/tls/generate_cert.go"
+					strs := []string{
+						"run",
+						path,
+						"--rsa-bits=2048",
+						"--host=localhos",
```
kevindiu, line 559:
```suggestion
						"--host=localhost",
```
```suggestion
						"--host=localhos",
```

internal/config/cassandra_test.go
```diff
@@ -296,128 +442,355 @@ func TestCassandra_Opts(t *testing.T) {
...
+					}
+					err := exec.Command("go", strs...).Run()
```
kevindiu, line 561: should we use another function to generate cert instead of using command?

kevindiu (commented):
I think we need author review for this PR.
Please add `author review` to the PR comment.

kevindiu (approved):
LGTM

internal/config/cassandra_test.go
```diff
@@ -296,128 +437,341 @@ func TestCassandra_Opts(t *testing.T) {
...
+					KVTable:                  "kv",
+					VKTable:                  "vk",
+					VectorBackupTable:        "backup_vector",
+				},
+				want: want{
+					wantOpts: make([]cassandra.Option, 0),
+					err:      errors.ErrInvalidDNSConfig(5*time.Minute, 1*time.Minute),
+				},
+			}
+		}()}
```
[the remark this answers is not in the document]
kevindiu: please fix it.

kevindiu (dismissed):
LGTM, please ask author to review it.

vankichi:
@kpango @rinx 
please review the changes at `internal/config/cassandra.go` 🙏

kpango (approved):
LGTM

internal/config/cassandra.go
```diff
@@ -209,13 +216,15 @@ func (cfg *Cassandra) Opts() (opts []cassandra.Option, err error) {
...
 	if cfg.Net != nil {
 		der, err := net.NewDialer(cfg.Net.Opts()...)
-		if err == nil {
-			opts = append(opts,
-				cassandra.WithDialer(
-					der,
-				),
-			)
+		if err != nil {
+			return nil, err
```
rinx, line 220: I'm not sure but is it okay to return `nil` here?
At L237, even if an error occurred at the line before it, it returns `opts`. What's the difference between them?

kevindiu: I think you are correct, but I think we need to return the options even when the error occurred.
e.g.
```go
	if cfg.Net != nil {
		der, err := net.NewDialer(cfg.Net.Opts()...)
                if err == nil {
                      opts = append(opts,
			cassandra.WithDialer(
				der,
			),
		)
            } else {
                   errs = errors.Wrap(errs, err.Err())
           }
       }

	if cfg.TLS != nil && cfg.TLS.Enabled {
		tcfg, err := tls.New(
			tls.WithCert(cfg.TLS.Cert),
			tls.WithKey(cfg.TLS.Key),
			tls.WithCa(cfg.TLS.CA),
		)
		if err != nil {
			return opts, errors.Wrap(errs, err)
		}
```

vankichi: @rinx thanks for the review.
Actually, I'm not sure whether which is best.
I just fix this along with `internal/config/mysql.go`.
https://github.com/vdaas/vald/blob/master/internal/config/mysql.go#L110-L113

But, if `net.newDialer()` returns the error, we cannot connect to the Cassandra cluster (is it right?).
If it is so, I think returning an error is not bad to prevent other errors, and throwing away the opts that are created before is too.

What do you think about this?
I think both are not bad and I'd like to get your suggestion as an author. 
cc: @kpango

rinx: To tell the truth, we don't have any obvious policy to handle problems like this, so i cannot answer it immediately.

> But, if net.newDialer() returns the error, we cannot connect to the Cassandra cluster (is it right?).

Even if it cannot get a `der` from `newDialer()`, internal/cassandra will work with the default dialer of gocql.

kevindiu: But even if `internal/cassandra` works with the default dialer, it may not work as excepted from the users...

rinx: Sure.
There's the same problem about L237.

So we should think about how to handle problems like this.

kpango: The previous implementations had a policy of implicitly using the Default Dialer without returning an error, but since it seems that implicit continuous should not be done, I thought it would be best practice for the API definition to return an error and ask the user for an appropriate error handle, which is why I LGTM'd this fix.

However, there has been no discussion on whether or not to return the return value of origin when returning an error.
Normally, it is better to return zero or empty value in case of an error, so I personally think it is better to follow this rule.

vankichi: ok, it seems we have to discuss this and get the way how we handle it.
Depending on the discussion, we have to update `internal/config/mysql.go#L237` too.

kevindiu: should we also return `nil, error` on line 238?
https://github.com/vdaas/vald/pull/1117/files#diff-82743e089bd2afaf22dfee471faeb927c8f7c87d9f392badc6c5d82431e83680L228

kevindiu: `However, there has been no discussion on whether or not to return the return value of origin when returning an error.`
I think we should discuss it.

vankichi: I think there are 3 options for this problem.
1. ignore the error
    - it does not seems good handling 😅 
2. return error but other functional options are still alive, and output Error (or Warn) Log.
    - if the error does not cause a critical issue on the Vald, it can be.
    - user will get the configurations error via each component's log and then can fix it.
3. return error and all of the options are restored.
    - if the error will have a fatal effect on running the Vald, we should stop the applying configuration.

Maybe we have to use properly option 2 and 3 based on the situation?
If it is yes, we need to define the policy.
What do you think?

kpango: 3

internal/config/cassandra_test.go
```diff
@@ -296,128 +437,342 @@ func TestCassandra_Opts(t *testing.T) {
...
+					DisableSchemaEvents:      false,
+					DisableSkipMetadata:      false,
+					DefaultIdempotence:       false,
+					WriteCoalesceWaitTime:    "200ms",
+					KVTable:                  "kv",
+					VKTable:                  "vk",
+					VectorBackupTable:        "backup_vector",
+				},
+				want: want{
+					wantOpts: make([]cassandra.Option, 0),
```
kevindiu, line 654:
```suggestion
					wantOpts: nil,
```

internal/config/cassandra_test.go
```diff
@@ -296,128 +437,342 @@ func TestCassandra_Opts(t *testing.T) {
...
+				},
+				want: want{
+					wantOpts: make([]cassandra.Option, 45),
+					err:      nil,
+				},
+			}
+		}(),
+		func() test {
+			return test{
+				name: "return 0 cassandra.Option and error when TLS config value in the invalid value",
```
kevindiu, line 560:
```suggestion
				name: "return nil and error when TLS config value in the invalid value",
```

internal/config/cassandra_test.go
```diff
@@ -296,128 +437,342 @@ func TestCassandra_Opts(t *testing.T) {
...
+					DisableSchemaEvents:      false,
+					DisableSkipMetadata:      false,
+					DefaultIdempotence:       false,
+					WriteCoalesceWaitTime:    "200ms",
+					KVTable:                  "kv",
+					VKTable:                  "vk",
+					VectorBackupTable:        "backup_vector",
+				},
+				want: want{
+					wantOpts: make([]cassandra.Option, 0),
```
hlts2, line 759: sorry, if I made a misstake
```suggestion
					wantOpts: nil,
```

internal/config/cassandra_test.go
```diff
@@ -296,128 +437,342 @@ func TestCassandra_Opts(t *testing.T) {
...
+					err:      errors.ErrTLSCertOrKeyNotFound,
+				},
+			}
+		}(),
+		func() test {
+			cert := testdata.GetTestdataPath("tls/dummyServer.crt")
+			key := testdata.GetTestdataPath("tls/dummyServer.key")
+			ca := testdata.GetTestdataPath("tls/dummyCa.pem")
+			return test{
+				name: "return 0 cassandra.Option and err when net.NewDialer returns error",
```
hlts2, line 664:
```suggestion
				name: "return nil and err when net.NewDialer returns error",
```

kevindiu (approved):
LGTM

hlts2 (approved):

kevindiu:
@kpango @rinx just in case, can you please review this PR?

rinx (approved):

kpango (approved):
LGTM

PR #1171, opened 2021-04-12
Implement safety bench code

kevindiu:
Signed-off-by: kevindiu <kevin_diu@yahoo.com.hk>

### Description:

This PR implements safety benchmark test.
Here is the benchmark result.
```
goos: darwin
goarch: amd64
pkg: github.com/vdaas/vald/internal/safety
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkRecoverFunc/fn_return_error_func-12         	134723181	         9.512 ns/op	      24 B/op	       1 allocs/op
BenchmarkRecoverWithoutPanicFunc/fn_return_error_func-12         	55744567	        19.84 ns/op	      40 B/op	       2 allocs/op
BenchmarkRecoverWithoutPanicFunc/fn_panic_runtime_error_func-12  	   38184	     30841 ns/op	   18041 B/op	     206 allocs/op
BenchmarkRecoverWithoutPanicFunc/fn_panic_string_func-12         	   38880	     31979 ns/op	   16919 B/op	     191 allocs/op
BenchmarkRecoverWithoutPanicFunc/fn_panic_error_func-12          	   36880	     33284 ns/op	   16869 B/op	     188 allocs/op
BenchmarkRecoverWithoutPanicFunc/fn_panic_int_func-12            	   36000	     36941 ns/op	   16902 B/op	     190 allocs/op
PASS
ok  	github.com/vdaas/vald/internal/safety	10.656s
```

After some try, seems the `TestMain()` in unit test will be executed before the benchmark function is executed, so I changed the `TestMain()` implementation to init `NOP` logger to avoid large amount of logging message to be printed.

Also there is a difference in `BenchmarkRecoverFunc()` and `BenchmarkRecoverWithoutPanicFunc()`.
Since `BenchmarkRecoverFunc()` will panic the runtime and it will stop the execution, the benchmark implementation of  `BenchmarkRecoverFunc()` will not execute the handler returned from the function.

### Environment:

- Go Version: 1.16
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/safety/safety_bench_test.go
+++ b/internal/safety/safety_bench_test.go
@@ -1,0 +1,140 @@
+package safety
+
+import (
+	"testing"
+
+	"github.com/vdaas/vald/internal/errors"
+)
+
+func BenchmarkRecoverFunc(b *testing.B) {
+	type args struct {
+		fn func() error
+	}
+	type test struct {
+		name string
+		args args
+	}
+	tests := []test{
+		{
+			name: "fn return error",
+			args: args{
+				fn: func() error {
+					return errors.New("fn err")
+				},
+			},
+		},
+	}
+	for _, test := range tests {
+		b.Run(test.name, func(b *testing.B) {
+			b.RunParallel(func(pb *testing.PB) {
+				for pb.Next() {
+					got := RecoverFunc(test.args.fn)
+					if got == nil {
+						b.Error("got is empty")
+					}
+				}
+			})
+		})
+	}
+}
+
+func BenchmarkRecoverWithoutPanicFunc(b *testing.B) {
+	type args struct {
+		fn func() error
+	}
+	type test struct {
+		name string
+		args args
+	}
+	tests := []test{
+		{
+			name: "fn return error",
+			args: args{
+				fn: func() error {
+					return errors.New("fn err")
+				},
+			},
+		},
+	}
+	for _, test := range tests {
+		b.Run(test.name, func(b *testing.B) {
+			b.RunParallel(func(pb *testing.PB) {
+				for pb.Next() {
+					got := RecoverWithoutPanicFunc(test.args.fn)
+					if got == nil {
+						b.Error("got is empty")
+					}
+				}
+			})
+		})
+	}
+}
+
+func BenchmarkRecoverWithoutPanicFunc_with_execution(b *testing.B) {
+	type args struct {
+		fn func() error
+	}
+	type test struct {
+		name string
+		args args
+	}
+	tests := []test{
+		{
+			name: "fn return error",
+			args: args{
+				fn: func() error {
+					return errors.New("fn err")
+				},
+			},
+		},
+		{
+			name: "fn panic runtime",
+			args: args{
+				fn: func() error {
+					_ = []string{}[10]
+					return nil
+				},
+			},
+		},
+		{
+			name: "fn panic string",
+			args: args{
+				fn: func() error {
+					panic("panic")
+				},
+			},
+		},
+		{
+			name: "fn panic error",
+			args: args{
+				fn: func() error {
+					panic(errors.Errorf("error"))
+				},
+			},
+		},
+		{
+			name: "fn panic int",
+			args: args{
+				fn: func() error {
+					panic(10)
+				},
+			},
+		},
+	}
+	for _, test := range tests {
+		b.Run(test.name, func(b *testing.B) {
+			b.RunParallel(func(pb *testing.PB) {
+				for pb.Next() {
+					got := RecoverWithoutPanicFunc(test.args.fn)
+					if got == nil {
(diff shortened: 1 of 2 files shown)


internal/safety/safety_bench_test.go
```diff
@@ -0,0 +1,140 @@
...
+	type args struct {
+		fn func() error
+	}
+	type test struct {
+		name string
+		args args
+	}
+	tests := []test{
+		{
+			name: "fn return error",
```
kevindiu, line 19: is it ok to name it `fn return error`?

internal/safety/safety_bench_test.go
```diff
@@ -0,0 +1,140 @@
...
+					if got == nil {
+						b.Error("got is empty")
+					}
+				}
+			})
+		})
+	}
+}
+
+func BenchmarkRecoverWithoutPanicFunc_with_execution(b *testing.B) {
```
kevindiu, line 73: is the benchmark name here ok?

internal/safety/safety_bench_test.go
hlts2: 🤔 
for example, how about `fn function that returns an error`?

internal/safety/safety_bench_test.go
```diff
@@ -0,0 +1,140 @@
...
+	type args struct {
+		fn func() error
+	}
+	type test struct {
+		name string
+		args args
+	}
+	tests := []test{
+		{
+			name: "fn return error",
```
hlts2, line 83: how about fn function that returns an error?

internal/safety/safety_bench_test.go
vankichi: `fn return error func`

internal/safety/safety_bench_test.go
vankichi: what's the meaning of `with execution`?

vankichi: we do not create such as agreement 🤔 
you should  name just `BenchmarkRecoverWithoutPanicFunc`
@hlts2 how do you think?

hlts2: > you should name just BenchmarkRecoverWithoutPanicFunc
@hlts2 how do you think?

I think it looks good to me 👍

internal/safety/safety_test.go
```diff
@@ -32,7 +32,7 @@ var goleakIgnoreOptions = []goleak.Option{
 }
 
 func TestMain(m *testing.M) {
-	log.Init()
+	log.Init(log.WithLoggerType("nop"))
```
hlts2, line 35: should we use string via const value to avoid corrections due to naming changes? 🤔 
```suggestion
	log.Init(log.WithLoggerType(logger.NOP.String()))
```

internal/safety/safety_bench_test.go
kevindiu: It conflicts with the benchmark function on line 41.

Why I named like this is because the `RecoverWithoutPanicFunc` return the function with panic handler, and this function benchmark the execution of returning function.
It is different with `BenchmarkRecoverWithoutPanicFunc` which only benchmark the `RecoverWithoutPanicFunc` function

internal/safety/safety_bench_test.go
kevindiu: @hlts2 what do you think about vankichi suggestion?

hlts2: @kevindiu 
sorry, it looks good to me 👍 
cc: @vankichi

internal/safety/safety_bench_test.go
hlts2: Got It!!
btw `BenchmarkRecoverWithoutPanicFunc` is a benchmark test that returns a function, and even if we get it, we can't get meaningful data, so how about putting it together?

hlts2: for example, It's like rewriting the contents of `BenchmarkRecoverWithoutPanicFunc` to the contents of `BenchmarkRecoverWithoutPanicFunc_with_execution`.
what do you think?

kevindiu: Sorry but I am not sure what do you mean by ` It's like rewriting the contents of BenchmarkRecoverWithoutPanicFunc to the contents of BenchmarkRecoverWithoutPanicFunc_with_execution.`

hlts2 (dismissed):
LGTM

vankichi (dismissed):

PR #1185, opened 2021-04-15
remove nil check in Error() method in internal/errors

vankichi:

### Description:

This PR includes:
- Remove nil check in `Error()` method in each `internal/errors` files.
- Add test for improve `internal/errors`'s test coverage

### Environment:

- Go Version: 1.16
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [ ] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -79,6 +79,11 @@ var (
 		return Wrapf(err, "failed to output %s logs", str)
 	}
 
+	// ErrErrorIsNil represents a function to generate an error that given name's error object is nil.
+	ErrErrorIsNil = func(name string) error {
+		return Errorf("error object is nil: %s", name)
+	}
+
 	// New represents a function to generate the new error with a message.
 	// When the message is nil, it will return nil instead of an error.
 	New = func(msg string) error {
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -175,8 +175,8 @@ var (
 	// As represents errors.As.
 	As = errors.As
 
-	// errNilCustomError represents a function to generate an error that given name's error object is nil.
-	errNilCustomError = func(name string) string {
-		return fmt.Sprintf("custom error object is nil: %s", name)
+	// errExpectedErrIsNil represents a function to generate an error that given name's error object is nil.
+	errExpectedErrIsNil = func(n string) error {
+		return Errorf("expected err is nil: %s", n)
 	}
 )
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -79,11 +79,6 @@ var (
 		return Wrapf(err, "failed to output %s logs", str)
 	}
 
-	// ErrErrorIsNil represents a function to generate an error that given name's error object is nil.
-	ErrErrorIsNil = func(name string) error {
-		return Errorf("error object is nil: %s", name)
-	}
-
 	// New represents a function to generate the new error with a message.
 	// When the message is nil, it will return nil instead of an error.
 	New = func(msg string) error {
@@ -179,4 +174,9 @@ var (
 
 	// As represents errors.As.
 	As = errors.As
+
+	// errNilCustomError represents a function to generate an error that given name's error object is nil.
+	errNilCustomError = func(name string) string {
+		return fmt.Sprintf("custom error object is nil: %s", name)
+	}
 )
--- a/internal/errors/ngt_test.go
+++ b/internal/errors/ngt_test.go
@@ -367,6 +367,79 @@ func TestErrDimensionLimitExceed(t *testing.T) {
 	}
 }
 
+func TestErrIncompatibleDimensionSize(t *testing.T) {
+	type args struct {
+		req int
+		dim int
+	}
+	type want struct {
+		want error
+	}
+	type test struct {
+		name       string
+		args       args
+		want       want
+		checkFunc  func(want, error) error
+		beforeFunc func(args)
+		afterFunc  func(args)
+	}
+	defaultCheckFunc := func(w want, got error) error {
+		if !Is(got, w.want) {
+			return Errorf("got: \"%#v\",\n\t\t\t\twant: \"%#v\"", got, w.want)
+		}
+		return nil
+	}
+	tests := []test{
+		{
+			name: "return an ErrIncompatibleDimensionSize error when req is 640 and dim is 720",
+			args: args{
+				req: 640,
+				dim: 720,
+			},
+			want: want{
+				want: New("incompatible dimension size detected\trequested: 640,\tconfigured: 720"),
+			},
+		},
+		{
+			name: "return an ErrIncompatibleDimensionSize error when req is empty and dim is 720",
+			args: args{
+				dim: 720,
+			},
+			want: want{
+				want: New("incompatible dimension size detected\trequested: 0,\tconfigured: 720"),
+			},
+		},
+		{
+			name: "return an ErrIncompatibleDimensionSize error when req is 640 and dim is 720",
+			args: args{
+				req: 640,
+			},
+			want: want{
+				want: New("incompatible dimension size detected\trequested: 640,\tconfigured: 0"),
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(tt *testing.T) {
+			if test.beforeFunc != nil {
+				test.beforeFunc(test.args)
+			}
+			if test.afterFunc != nil {
+				defer test.afterFunc(test.args)
+			}
+			if test.checkFunc == nil {
+				test.checkFunc = defaultCheckFunc
+			}
+
+			got := ErrIncompatibleDimensionSize(test.args.req, test.args.dim)
+			if err := test.checkFunc(test.want, got); err != nil {
+				tt.Errorf("error = %v", err)
+			}
+		})
+	}
+}
+
 func TestErrUnsupportedObjectType(t *testing.T) {
 	type want struct {
 		want error
--- a/internal/errors/ngt_test.go
+++ b/internal/errors/ngt_test.go
@@ -410,7 +410,7 @@ func TestErrIncompatibleDimensionSize(t *testing.T) {
 			},
 		},
 		{
-			name: "return an ErrIncompatibleDimensionSize error when req is 640 and dim is 720",
+			name: "return an ErrIncompatibleDimensionSize error when req is 640"
 			args: args{
 				req: 640,
 			},
--- a/internal/errors/ngt_test.go
+++ b/internal/errors/ngt_test.go
@@ -410,7 +410,7 @@ func TestErrIncompatibleDimensionSize(t *testing.T) {
 			},
 		},
 		{
-			name: "return an ErrIncompatibleDimensionSize error when req is 640"
+			name: "return an ErrIncompatibleDimensionSize error when req is 640",
 			args: args{
 				req: 640,
 			},
--- a/internal/errors/redis.go
+++ b/internal/errors/redis.go
@@ -73,10 +73,7 @@ type ErrRedisNotFoundIdentity struct {
 
 // Error returns the string of ErrRedisNotFoundIdentity.error.
 func (e *ErrRedisNotFoundIdentity) Error() string {
-	if e.err != nil {
-		return e.err.Error()
-	}
-	return ""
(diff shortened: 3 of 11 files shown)


internal/errors/redis.go
```diff
@@ -73,10 +73,7 @@ type ErrRedisNotFoundIdentity struct {
 
 // Error returns the string of ErrRedisNotFoundIdentity.error.
 func (e *ErrRedisNotFoundIdentity) Error() string {
-	if e.err != nil {
-		return e.err.Error()
-	}
-	return ""
+	return e.err.Error()
```
hlts2, line 76: What about the previous author's feedback as to whether the error should be exported? :thinking:

vankichi: I think it should return an error string as same as golang's default  `func(e *error) Error() string{}`.

hlts2: I'm sorry, my way of communicating was not good. 
I wanted to tell you whether it should be the following A implementation(exported error) or the following B(Unexported error) implementation.

A. (Exported error)
```go
type ErrRedisNotFoundIdentity struct {
  Err error <- this
}

func (e *ErrRedisNotFoundIdentity) Error() string {
...
}
```


B. (Unexported error)
```go
type ErrRedisNotFoundIdentity struct {
  errr error <- this
}

func (e *ErrRedisNotFoundIdentity) Error() string {
...
}
```

vankichi: Nop. I think we should not return `" ` as the returned value.
It is acceptable to cause nil pointer error.
But, I think each custom error struct has implemented correctly, nil pointer error will not occur.

hlts2 (approved):
LGTM

internal/errors/redis.go
kevindiu: @hlts2 can you tell us what is the advantages of exporting the internal error?

kevindiu (approved):
LGTM

kevindiu:
do we need author review for this PR?

kevindiu:
@vankichi

internal/errors/redis.go
```diff
@@ -73,10 +73,7 @@ type ErrRedisNotFoundIdentity struct {
 
 // Error returns the string of ErrRedisNotFoundIdentity.error.
 func (e *ErrRedisNotFoundIdentity) Error() string {
-	if e.err != nil {
-		return e.err.Error()
```
kevindiu, line 77: like kpango said, should we define an error message for `e.err` is nil?

vankichi: How about defining a new error function in errors.go like as 
```go
ErrErrorIsNil = func(name string) {
    return Errorf("error object is nil: %s". name)
}
```

and then implement in custom error struct:

```go
func (e *customError) Error() string {
    if e.err == nil {
        return ErrErrorIsNil(reflect.typeOf(e).Elem().Name()).Error()
    }
    return e.err.Error()
}
```

any ideas?

kevindiu: I think we can remove the internal `err` from the field in this situation. e.g. on line 28:
Current implementation:
```golang
	// NewErrRedisNotFoundIdentity represents a function to generate an ErrRedisNotFoundIdentity error.
	NewErrRedisNotFoundIdentity = func() error {
		return &ErrRedisNotFoundIdentity{
			err: New("error redis entry not found"),
		}
	}
```

suggestion:
```golang
	// NewErrRedisNotFoundIdentity represents a function to generate an ErrRedisNotFoundIdentity error.
	NewErrRedisNotFoundIdentity = func() error {
		return  New("error redis entry not found")
	}
```

vankichi: Thx. I'd like to confirm your suggestion.
1. Your suggestion is to change the struct. Why dir you think like this?
1. have you confirmed the impact of this change?
1. you may have to consider other implements. How do you think about this?

please check these questions. @kevindiu 
@hlts2 how do you think about those discussions? and please share your idea.

kevindiu: @vankichi Why I think this is because the `err` field is not using. 
It cannot access outside this package, so I think we can safely remove it.

I have checked the implementation of  `NewErrRedisNotFoundIdentity` and seems no one is using it, so I think it is safe to make this changes to it.

Consider other implementation, I think your suggestion is fine.

vankichi: I meant did you check e.g. `internal/errors/option`? @kevindiu

internal/errors/redis.go
hlts2: sorry to late reply.
At the time of the oral discussion, it was said that there was no problem with the current policy, so I will close the comment.

internal/errors/redis.go
kevindiu: yes, I think we cannot apply this on errors/option.

but in today's meeting, PO suggested to define a string type contains the default error string and then return it.
I think it is the easiest way to fix it.

vankichi: So, is it ok?
```go
func (e *customError) Error() string {
    if e.err == nil {
	e.err = New("error redis entry not found")
    }
    return e.err.Error()
}
```

kevindiu: I think it should be fine.
```go
var (
   defaultCustomErrorString = "error ....."
)

func (e *customError) Error() string {
    if e.err == nil {
	return defaultCustomErrorString
    }
    return e.err.Error()
}
``

Comment (vankichi) [reply internal/errors/redis.go:77]: well, so could you give me your idea about the `derfaultCustomErrorString`.?

Comment (kevindiu) [reply internal/errors/redis.go:77]: the idea is to return a default string when `err` is nil, without creating error type as you suggested.
I think it is the most simplest way and clean way to resolve this issue.

Comment (vankichi) [reply internal/errors/redis.go:77]: sorry, what I wanted to know is that can you show me the string example that is defined as `defaultCustomErrorString`

Comment (kevindiu) [reply internal/errors/redis.go:77]: sorry I misunderstood your question.
it is the same as you suggested. 
the variable name and the string value will be changed according to the target error type.

for instance, for the error `ErrRedisNotFoundIdentity`, 
```go
var (
   defaultErrRedisNotFoundIdentityString = "error redis entry not found"
)
```

vankichi: @kevindiu 
It is good when you have to think about only this case.
IMO, we should define the default error string variable that can cover all custom errors.

kevindiu: @vankichi thank you for your comment.
To simplify the implementation, do you think the below suggestion is good?
```go
func nilErrString(err string) string {
     return fmt.Sprintf("error object is nil: %s", err)
}
```

vankichi: @kevindiu 
Looks better!
maybe not needed the input value :thinking:

kevindiu: @vankichi Sorry my suggestion is not clear. maybe we need to know which error contains nil error.
To use this function, we pass the error name to this function. e.g. 
```go
func (e *ErrRedisNotFoundIdentity) Error() string {
    if e.err == nil {
        return nilErrString("ErrRedisNotFoundIdentity")
    }
    return e.err.Error()
}
```

vankichi: the log will show the point where the error occurs, so we do not have to use any input value to specify which custom error causes the error.

kevindiu: Do you mean by seeing the stack trace? 
I think sometimes the log only shows `[ERR] xxxxxx` and the cause of the error may not be visible.

kevindiu: Please let me know if I misunderstood, thank you :)

kevindiu: @vankichi remind :)

vankichi: how is it?
https://github.com/vdaas/vald/commit/5c7fd9cb3c6997b5e1d3ba2e91d4154bfa0cf81f

kevindiu: @vankichi I think it is good.
cc: @hlts2 what do you think?

internal/errors/errors.go
```diff
@@ -174,4 +174,9 @@ var (
 
 	// As represents errors.As.
 	As = errors.As
+
+	// errNilCustomError represents a function to generate an error that given name's error object is nil.
+	errNilCustomError = func(name string) string {
```
hlts2, line 179: Should we define pure function instead of function variable?:thinking:

vankichi: Up to us.

hlts2: got it.

hlts2 (dismissed):

kevindiu (approved):
LGTM

internal/errors/errors.go
kevindiu, line 179: Sorry again but kpango has asked if the following suggestion is better.
```suggestion
	errNilCustomError = func(name string) error {
```
@vankichi @hlts2 what do you think??

hlts2: I think both are good for me.
But In terms of how to create an error package, I think it is better to return an error.
and  should we change from unexported value to exported value?

```
errNilCustomError = func(name string) string {
↓
ErrNilCustomError = func(name string) error {
```

kpango: First of all, I think the name ErrNilCustomError is not clear.
I recommend `ErrExpectedErrorIsNil`

kevindiu: @hlts2 I agree with you that it's in error package and it may return an error.
But I think this function should be unexported because it shouldn't be used outside this package.

vankichi: @kevindiu @hlts2 @kpango 
```go
ErrExpectedErrIsNil(n string) error {
    return Errorf("expected err is nil: %s", n)
}
```
How is it?

kevindiu: I think it should be `errExpectedErrIsNil` instead of `ErrExpectedErrIsNil` 🤔

vankichi: @kevindiu why do you think?
@hlts2 why do you think it should be exported value? I guessed this will be used only internal errors package.

hlts2: @vankichi 
>
```
ErrExpectedErrIsNil = func(n string) error {
    return Errorf("expected err is nil: %s", n)
}
```How is it?

it looks good to me

Comment (kpango) [reply internal/errors/errors.go:179]: I think it's getting better, but we need to be concerned about @kevindiu's opinion on whether we should publish this variable or not.

Comment (kevindiu) [reply internal/errors/errors.go:179]: @vankichi Like I said, I think this function should be unexported because it shouldn't be used outside this package.
What do you think?

Comment (kevindiu) [reply internal/errors/errors.go:179]: @hlts2 to define it is unexported function so that it cannot be used outside this package

Comment (hlts2) [reply internal/errors/errors.go:179]: I got it. thank you.
if this function is not used outside, I think the unexported function is better.

Comment (vankichi) [reply internal/errors/errors.go:179]: so we define `errExpectedErrIsNil` instead of `ErrExpectedErrIsNIl`

Comment (kevindiu) [reply internal/errors/errors.go:179]: thank you!

File: internal/errors/ngt_test.go [hunk 4]
```diff
@@ -367,6 +367,79 @@ func TestErrDimensionLimitExceed(t *testing.T) {
...
+			name: "return an ErrIncompatibleDimensionSize error when req is empty and dim is 720",
+			args: args{
+				dim: 720,
+			},
+			want: want{
+				want: New("incompatible dimension size detected\trequested: 0,\tconfigured: 720"),
+			},
+		},
+		{
+			name: "return an ErrIncompatibleDimensionSize error when req is 640 and dim is 720",
```
kevindiu, line 413:
```suggestion
			name: "return an ErrIncompatibleDimensionSize error when req is 640"
```

kevindiu (changes requested):

hlts2 (dismissed):

kevindiu (approved):
LGTM

vankichi:
I'll fix soon

vankichi:
I should check carefully 😅 
https://github.com/vdaas/vald/pull/1185#discussion_r630888296

kevindiu:
sorry we should also check it carefully. 😢

PR #1195, opened 2021-04-22
Add general info metrics package / Add NGT info metrics

rinx:
### Description:

This PR includes two main features.
- Add general info metrics package
    - enables us to add new info metrics easily, based on internal/config structs.
    - added unit tests. already passed with 100% coverage.
- Add NGT info metrics
    - includes "distance type", "object type" and "dimension".
    - using Prometheus exporter, the metrics looks like:
```
# HELP vald_agent_core_ngt_info Agent NGT info
# TYPE vald_agent_core_ngt_info gauge
vald_agent_core_ngt_info{dimension="784",distance_type="l2",object_type="float"} 1
```

### Related Issue:

Nothing

tested on my k3d cluster.

### Environment:

- Go Version: 1.16
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


internal/observability/metrics/info/info.go
```diff
@@ -0,0 +1,138 @@
...
+		default:
+			continue
+		}
+
+		k, err := metrics.NewKey(keyName)
+		if err != nil {
+			return nil, err
+		}
+
+		kvs[k] = value
```
kpango, line 104: 多分、こんな感じ
```suggestion
                k, err := metrics.NewKey(keyName)
                if err != nil {
                        return nil, err
                }
                switch v := rv.Field(k); v.Kind() {
                case reflect.String:
                        kvs[k] = v.String()
                case reflect.Slice, reflect.Array:
                        kvs[k] = fmt.Sprintf("%.255s", fmt.Sprintf("%v", v.Interface()))
                case reflect.Bool:
                        kvs[k] = strconv.FormatBool(v.Bool())
                case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
                        kvs[k] = strconv.FormatInt(v.Int(), 10)
                case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
                        kvs[k] = strconv.FormatUint(v.Uint(), 10)
                case reflect.Float32, reflect.Float64:
                        kvs[k] = strconv.FormatFloat(v.Float(), 'E', -1, 64)
                }

```

kpango (changes requested):
Others LGTM

internal/observability/metrics/info/info.go
rinx: thanks, but if `v` is something other than primitive type or slice of a primitive type (maybe struct or slice of a struct), it should be ignored.

kpango: sorry what do you mean???
for slice value???

rinx: the input value of this function `i` can be any struct type like
```
type A struct {
	B string
	C struct {...}
	D []struct {...}
}
```
.
In this case, C and D should be ignored. `metrics.KeyName` should not be called with these `keyName`s.
In the previous code, `default` branch is for this purpose.

So I think, 
```
                case reflect.Slice, reflect.Array:
```
should be modified to accept only `[]string`.

rinx: Technically it is possible to put complicated struct values into the label. However it is not a good manner for metrics labeling.

kpango: I've fixed the string slice.

However, I haven't set it to kvs for the struct type, but will it affect to the label?

kpango: I noticed that the value of k has been overwritten, you should use the alias of the key used in the for loop

rinx: thanks. I'll revise it.

Just calling `metrics.NewKey(keyName)` makes a new label in the global instance. It is not a good behavior.

kpango: hmm, okay I'll revise it after switch

rinx: there's no need to revise. I'll revise this PR following your code. :)

kpango: I already updated

rinx: revised.

rinx: Is there something wrong? :thinking:

kpango: seems good I'm checking.

rinx: Sorry, I've just added several missing test cases. Please check it :pray:

kpango (approved):
LGTM

PR #1231, opened 2021-05-14
Bug fix of parallel test and goleak

hlts2:

### Description:

The cause is that the top-level test functions are also parallelized and could not be handled properly by goleak.CurrentIgnore.
In the current code, the top-level test functions and each test case run in parallel.
For example, if TestA runs and `goleak.IgnoreCurrent` in its internal test case, the goroutine at that moment will be ignored. However, if TestB is executed in parallel immediately after goleak.IgnoreCurrent, it will increase from the number of goruotines at the time of TestA. That is the cause of this time.
When using goleak in test, the conclusion is that each test case can be run in parallel, but top-level test functions cannot be run in parallel.

### Environment:

- Go Version: 1.16
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [x] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/assets/test/templates/common/function.tmpl
+++ b/assets/test/templates/common/function.tmpl
@@ -2,7 +2,6 @@
 {{- $f := . }}
 
 func {{ .TestName }}(t *testing.T) {
-    t.Parallel()
     {{- if .TestParameters }}
     type args struct {
         {{- range .TestParameters }}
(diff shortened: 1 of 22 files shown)


kpango (approved):
LGTM

PR #1288, opened 2021-06-15
Bugfix index count problem of ngt

hlts2:
### Description:

In this PR, we fixed the following problems in vald agent:
1. When `saveIndex()` runs, the kvs index count and metadata.json index count may be different due to the timing of executing the `kvs.Len()`. We fixed to reference the kvs len variable to the same variable.
2. When creating index runs, the timing of removing from the kvs and the internal vqueue is different, which may cause the inconsistency problem of kvs&vqueue. We changed by passing the callback function to the vqueue to remove the kvs&vqueue data at the same time to avoid the above problem.
3. If the vector is updated during delete phase of createIndex operation, the vector will be inserted in insert phase of createIndex operation and the vector will be removed in the next createIndex operation.

### Environment:

- Go Version: 1.16.4
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [x] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/pkg/agent/core/ngt/service/vqueue/queue.go
+++ b/pkg/agent/core/ngt/service/vqueue/queue.go
@@ -309,29 +309,22 @@ func (v *vqueue) flushAndRangeDelete(f func(uuid string) bool) {
 		return udk[i].date > udk[j].date
 	})
 	dup := make(map[string]bool, len(udk)/2)
-	dl := make([]int, 0, len(udk)/2)
+	udm := make(map[string]int64, len(udk))
 	for i, idx := range udk {
-		if dup[idx.uuid] {
-			dl = append(dl, i)
-		} else {
+		if !dup[idx.uuid] {
 			dup[idx.uuid] = true
+			if !f(idx.uuid) {
+				v.dmu.Lock()
+				v.udk = append(udk[i:], v.udk...)
+				v.dmu.Unlock()
+				return
+			}
 			v.udim.Delete(idx.uuid)
-			f(idx.uuid)
+			udm[idx.uuid] = idx.date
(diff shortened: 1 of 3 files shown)


pkg/agent/core/ngt/service/vqueue/queue.go
```diff
@@ -291,9 +273,9 @@ func (v *vqueue) addDelete(d key) {
 	v.dmu.Unlock()
 }
 
-func (v *vqueue) flushAndLoadInsert() (uii []index) {
+func (v *vqueue) flushAndRangeInsert(f func(uuid string, vector []float32) bool) {
```
hlts2, line 276: Should we pass `context.Context` ?
And if the context is canceled, should we return in L.295?

kevindiu:
Question: maybe we don't need channel(ich/dch) in vqueue and use the map (uiim/udim) directly?
https://github.com/vdaas/vald/blob/f7bb9f8faf107f666db39593aeaabc3bb73f7313/pkg/agent/core/ngt/service/vqueue/queue.go#L171

pkg/agent/core/ngt/service/vqueue/queue.go
kevindiu: I think we shouldn't stop the createIndex process when the context is canceled.
It is because if we stop createIndex, the index may be lost due to uncompleted process of vqueue.

kpango (approved):
LGTM

PR #1309, opened 2021-06-23
bugfix correct error backoff handling for lb gateway

kpango:

### Description:
I fixed correct error backoff handling for lb gateway

1. old status is below
LBGW never backoff on current implementation in all case,  it was not working correctly.
2. the bug detail is below.
In the gRPC Handler of LBGW, when BroadCast or DoMulti is executed, if it does not return an Error, the backoff of internal/net/grpc/client will not work, but LBGW does not return an Error using return nil.
3. how to fixed.
LBGW now returns the correct error to internal/net/grpc/client, but this client also backs off in some cases, such as ContextCancel. To address this issue, we have improved gRPC Status and fixed the backoff condition.

### Environment:

- Go Version: 1.16.4
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [x] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/pkg/gateway/lb/handler/grpc/handler.go
+++ b/pkg/gateway/lb/handler/grpc/handler.go
@@ -90,7 +90,27 @@ func (s *server) Exists(ctx context.Context, meta *payload.Object_ID) (id *paylo
 			oid, err := vc.Exists(sctx, &payload.Object_ID{
 				Id: meta.GetId(),
 			}, copts...)
-			if err != nil && !errors.Is(err, context.Canceled) {
+			if err != nil {
+				switch {
+				case errors.Is(err, context.Canceled),
+					errors.Is(err, errors.ErrRPCCallFailed(target, context.Canceled)):
+					if sspan != nil {
+						sspan.SetStatus(trace.StatusCodeCancelled(
+							errdetails.ValdGRPCResourceTypePrefix +
+								"/vald.v1.Exists.BroadCast/" +
+								target + " canceled: " + err.Error()))
+					}
+					return nil
+				case errors.Is(err, context.DeadlineExceeded),
+					errors.Is(err, errors.ErrRPCCallFailed(target, context.DeadlineExceeded)):
+					if sspan != nil {
+						sspan.SetStatus(trace.StatusCodeDeadlineExceeded(
+							errdetails.ValdGRPCResourceTypePrefix +
+								"/vald.v1.Exists.BroadCast/" +
+								target + " deadline_exceeded: " + err.Error()))
+					}
+					return nil
+				}
 				var (
 					st  *status.Status
 					msg string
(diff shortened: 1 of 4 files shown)


vankichi (dismissed):

rinx:
What was the problem? Please describe the bug.

kpango:
@rinx the problem is LBGateway never backoff on current implementation.
so I fixed `correct error backoff handling for lb gateway`

rinx:
I already know about that.
I don't understand what case doesn't it work correctly.

kpango:
In all cases, it was not working correctly.

kpango:
@rinx FYI
In the gRPC Handler of LBGW, when BroadCast or DoMulti is executed, if it does not return an Error, the backoff of internal/net/grpc/client will not work, but LBGW does not return an Error using return nil.

kpango:
I updated PR description please check it.

rinx:
Thanks. I finally understand the point to review.

pkg/gateway/lb/handler/grpc/handler.go
```diff
@@ -105,7 +109,7 @@ func (s *server) Exists(ctx context.Context, meta *payload.Object_ID) (id *paylo
 				if sspan != nil {
 					sspan.SetStatus(trace.FromGRPCStatus(st.Code(), msg))
 				}
-				return nil
+				return err
```
rinx, line 112: I'm not sure but I think it should not return an error if it is a NotFound error.

pkg/gateway/lb/handler/grpc/handler.go
```diff
@@ -888,7 +893,7 @@ func (s *server) Insert(ctx context.Context, req *payload.Insert_Request) (ce *p
 				}
 				emu.Unlock()
 			}
-			return nil
+			return err
```
rinx, line 896: I think it should not return an error if it is an AlreadyExists error.

pkg/gateway/lb/handler/grpc/handler.go
```diff
@@ -2098,7 +2103,7 @@ func (s *server) GetObject(ctx context.Context, req *payload.Object_VectorReques
 				if span != nil {
 					span.SetStatus(trace.FromGRPCStatus(st.Code(), msg))
 				}
-				return nil
+				return err
```
rinx, line 2106: Same. A NotFound error is ignorable.

rinx (commented):
Thanks for fixing the bug.
I added several comments but I'm not sure they are correct.

kpango:
@rinx thank you for reviewing.
I fixed the point that you pointed out.

FYI: vald's internal grpc client only backoffs when status is Internal, Resource-Exhausted, Unavailable.
https://github.com/vdaas/vald/blob/master/internal/net/grpc/client.go#L606-L611

vankichi (approved):

rinx (approved):

PR #1335, opened 2021-07-07
[bugfix] nil pointer panic in agent's MultiUpsert operation

kpango:

### Description:
In some cases, the MultiUpserAPI only executes either MultiUpdate or MultiInsert depending on the index status, and in those cases, the Location is nil, causing panic.
In this PullRequest, in addition to fixing the above issues, the following issues are also mentioned and fixed.
New issue: In MultiUpsert, if only MultiUpdate or MultiInsert is used, it is better to execute without dispatching goroutine to avoid the possibility of panic and to be more efficient in terms of CPU resources. This is clear.

In this PR I have modified the following based on the problems mentioned above.
In the case of MultiUpsert on Gateway and Agent, the number of executions of MultiInsert and MultiUpdate should be checked, and if one of them is less than 0, only the other one should be executed without using the goroutine, and it should be treated as the result of MultiUpsert.

Also, the Search API was returning ResourceExhausted when the CreateIndex was in progress, but this was considered to be incorrect because the API was cancelling the request for their own reasons, not because the resource was exhausted or had reached its limit, so it should be Aborted.
I also considered using "Canceled", but I thought "Canceled" would rather mean that the request was actively canceled by the client, so I decided to use "Aborted" to mean that the request was canceled from the Server.

* Please ignore the ResourceName fixes in grpc ResourceInfo status message as it is just a tiny refactoring

### Environment:

- Go Version: 1.16.5
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [x] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/pkg/gateway/lb/handler/grpc/handler.go
+++ b/pkg/gateway/lb/handler/grpc/handler.go
@@ -1637,7 +1637,7 @@ func (s *server) StreamUpsert(stream vald.Upsert_StreamUpsertServer) (err error)
 	return nil
 }
 
-func (s *server) MultiUpsert(ctx context.Context, reqs *payload.Upsert_MultiRequest) (locs *payload.Object_Locations, err error) {
+func (s *server) MultiUpsert(ctx context.Context, reqs *payload.Upsert_MultiRequest) (res *payload.Object_Locations, err error) {
 	ctx, span := trace.StartSpan(ctx, apiName+".MultiUpsert")
 	defer func() {
 		if span != nil {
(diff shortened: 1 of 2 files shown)


rinx (approved):

kevindiu (approved):

pkg/gateway/lb/handler/grpc/handler.go
```diff
@@ -1713,81 +1713,81 @@ func (s *server) MultiUpsert(ctx context.Context, reqs *payload.Upsert_MultiRequ
...
+		})
+	default:
+		var (
+			ures, ires *payload.Object_Locations
+			errs       error
+			mu         sync.Mutex
+		)
+		eg, ectx := errgroup.New(ctx)
+		eg.Go(safety.RecoverFunc(func() (err error) {
+			ures, err = s.MultiUpdate(ectx, &payload.Update_MultiRequest{
```
hlts2, line 1733: I'm not sure, but can the results be affected when calling MultiUpdate and MultiUpsert in parallel? 🤔

kpango: I didn't call MultiUpdate and MultiUpsert in parallel.

hlts2: I am sorry. I made a mistake. :bow:
It is about calling MultiUpdate and MultiInsert in parallel as shown below.
https://github.com/vdaas/vald/blob/fafd417d39444b24dda21b5465cc2c935f29d909/pkg/gateway/lb/handler/grpc/handler.go#L1732-L1761

kpango: I am not sure what your concern is.
However, what I am interested in is what problems you imagine would occur if MultiUpdate and MultiInsert were called in parallel.

hlts2: I understood. I read it again, so I realized that I made a mistake. sorry.
There is no problem with your implementation. I will close this comment.

hlts2 (approved):
LGTM

PR #1343, opened 2021-07-13
[bugfix] agent createindex operation's time.Ticker purges too slow & buffer overflow due to the unnecessary error wrapping

kpango:

### Description:
In the case of the CreateIndex, there is a possibility that the CreateIndex will not be executed while the SaveIndex is in progress, so the for-loop method should be used for the CreateIndex as well as the SaveIndex, and the Uncommitted Index Length should be checked before and after the CreateIndex.

Also, the CreateIndex, SaveIndex, and vqueue RangePop* functions run for a much longer time than the other API execution, but since the internal time.Ticker is terminated by time.After or derfer tick.Stop, so this PR fixes this as well.

### Environment:

- Go Version: 1.16.5
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/pkg/agent/core/ngt/service/ngt.go
+++ b/pkg/agent/core/ngt/service/ngt.go
@@ -592,20 +592,36 @@ func (n *ngt) CreateIndex(ctx context.Context, poolSize uint32) (err error) {
 			span.End()
 		}
 	}()
-
-	n.cimu.Lock()
-	defer n.cimu.Unlock()
-	if n.IsIndexing() || n.IsSaving() {
-		return nil
-	}
 	ic := n.vq.IVQLen() + n.vq.DVQLen()
 	if ic == 0 {
 		return errors.ErrUncommittedIndexNotFound
 	}
+	err = func() error {
+		ticker := time.NewTicker(time.Millisecond * 100)
+		defer ticker.Stop()
+		// wait for not indexing & not saving
+		for n.IsIndexing() || n.IsSaving() {
+			runtime.Gosched()
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-ticker.C:
+			}
+		}
+		return nil
+	}()
+	if err != nil {
+		return err
+	}
 	n.indexing.Store(true)
 	defer n.indexing.Store(false)
+	n.cimu.Lock()
+	defer n.cimu.Unlock()
 	defer n.gc()
-
+	ic = n.vq.IVQLen() + n.vq.DVQLen()
+	if ic == 0 {
+		return errors.ErrUncommittedIndexNotFound
+	}
 	log.Infof("create index operation started, uncommitted indexes = %d", ic)
 	log.Debug("create index delete phase started")
 	n.vq.RangePopDelete(ctx, func(uuid string) bool {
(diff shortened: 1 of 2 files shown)


pkg/agent/core/ngt/service/ngt.go
```diff
@@ -592,20 +592,36 @@ func (n *ngt) CreateIndex(ctx context.Context, poolSize uint32) (err error) {
...
+			}
+		}
+		return nil
+	}()
+	if err != nil {
+		return err
+	}
 	n.indexing.Store(true)
 	defer n.indexing.Store(false)
+	n.cimu.Lock()
```
kevindiu, line 618: I think this lock should place before line 616

pkg/agent/core/ngt/service/ngt.go
```diff
@@ -670,16 +682,22 @@ func (n *ngt) saveIndex(ctx context.Context) (err error) {
...
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-ticker.C:
+			}
 		}
+		return nil
+	}()
+	if err != nil {
+		return err
```
kevindiu, line 700: should we put lock here also?

pkg/agent/core/ngt/service/ngt.go
```diff
@@ -592,44 +592,56 @@ func (n *ngt) CreateIndex(ctx context.Context, poolSize uint32) (err error) {
...
-	}
 	ic := n.vq.IVQLen() + n.vq.DVQLen()
 	if ic == 0 {
 		return errors.ErrUncommittedIndexNotFound
 	}
+	err = func() error {
+		ticker := time.NewTicker(time.Millisecond * 100)
+		defer ticker.Stop()
+		// wait for not indexing & not saving
+		for n.IsIndexing() || n.IsSaving() {
```
kevindiu, line 603: Sorry that if I am wrong, but I think we can put this checking inside the lock without handling by ourself, or we can remove this checking since only 1 goroutine can enter the mutex lock scope. Other than ctx.Done handling, is there any other reason to write this logic here?

kpango: Since the CreateIndex and SaveIndex are called randomly and asynchronously via the WebAPI, it is necessary to check the Atomic Flag.
Also,`cimu` is a Lock for CreateIndex only, it is not affected by SaveIndex.
Rather, I feel that mutex is inherently more unnecessary.

pkg/agent/core/ngt/service/ngt.go
kpango: why???

kevindiu: same as createIndex, to avoid multiple goroutine to execute saveIndex simultaneously

hlts2 (approved):
thank you. It looks good to me!

pkg/agent/core/ngt/service/ngt.go
hlts2: @kevindiu 
To avoid that, I think `saveIndex` has the following code.
If called multiple times in multiple goroutine, `noice` will be the same as`lastNoice` except for the first time, so `saveIndex` may not be executed.

https://github.com/vdaas/vald/blob/b9e90f944715c488f76daf1a7878f08303835863/pkg/agent/core/ngt/service/ngt.go#L680-L684

kevindiu (approved):

PR #1353, opened 2021-07-20
Add E2E tests for multi-APIs

rinx:

### Description:

This PR contains:

- Add E2E tests for multi-APIs
- Use t.CleanUp for calling teardown
- Split operation.go into several files

### Related Issue:

nothing

nothing

### Environment:

- Go Version: 1.16.6
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [ ] Bug fix [type/bug]
- [X] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


kpango (dismissed):
LGTM

tests/e2e/operation/multi.go
```diff
@@ -0,0 +1,219 @@
...
+		})
+	}
+
+	req := &payload.Search_MultiRequest{
+		Requests: reqs,
+	}
+	_, err = client.MultiSearch(ctx, req)
+	if err != nil {
+		return err
+	}
```
kevindiu, line 54: I suggest to check the result length and also the distance contains 0 (exact result). 
Same as below functions.
```suggestion
	res, err = client.MultiSearch(ctx, req)
	if err != nil {
		return err
	}
	if len(res.Results) != len(ds.Test) {
	        return errors.New("response length does not match with dataset")
	}
	if res. Results[0].Distance != 0 {
	        return errors.New("first result is not 0")
	}
```

rinx: Thanks, but it should be like the following.
```suggestion
	res, err := client.MultiSearch(ctx, req)
	if err != nil {
		return err
	}
	
	if len(res.GetResponses()) != len(ds.Test) {
		return errors.New("number of responses does not match with sent requests")
	}
```

The first element of each result doesn't always have 0 distance.

kevindiu: thank you. should we check the result contains 0 distance element?

rinx: No. in almost cases, the results don't contain 0 distance element.

tests/e2e/operation/multi.go
```diff
@@ -0,0 +1,223 @@
...
+		reqs = append(reqs, &payload.Search_IDRequest{
+			Id:     strconv.Itoa(i),
+			Config: cfg,
+		})
+	}
+
+	req := &payload.Search_MultiIDRequest{
+		Requests: reqs,
+	}
+	_, err = client.MultiSearchByID(ctx, req)
```
kevindiu, line 86: should we check the length also? (all of the functions in this file)

kevindiu: other than this comment, LGTM for me

rinx: Thanks! I added them.

kevindiu (approved):
LGTM! thank you!

hlts2 (approved):
LGTM

PR #1398, opened 2021-08-16
remove global errgroup from kvsdb goroutine limitation

kpango:
### Description:
This bug was found and reported by @rinx.
In PR #1384, it was reported that the Agent NGT Service does not start (blocking) only when KVSDB.Concurrency is set to 0.
It is not recommended to use your own errgroup in kvsdb from cancellation point of view.
However, since the number of goroutines to be invoked remains the same, we believe that using our own errgroup is not a problem from a performance point of view.
Therefore, in this PR, KVSDB will keep its own errgroup and use it for concurrent map search.

If the concurrency is less than or equal to 0, goroutines will be created without any limitation, and if the concurrency is greater than or equal to 1, the search will be performed using the number of goroutines.

### Environment:

- Go Version: 1.16.6
- Docker Version: 19.03.8
- Kubernetes Version: 1.18.2
- NGT Version: 1.12.3

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/pkg/agent/core/ngt/service/kvs/kvs.go
+++ b/pkg/agent/core/ngt/service/kvs/kvs.go
@@ -66,16 +66,17 @@ func New(opts ...Option) BidiMap {
 	}
 	for i := range b.ou {
 		b.ou[i] = new(ou)
-	}
-	for i := range b.uo {
 		b.uo[i] = new(uo)
 	}
+
 	if b.eg == nil {
 		b.eg, _ = errgroup.New(context.Background())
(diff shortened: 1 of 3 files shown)


kevindiu (approved):
LGTM

rinx (approved):
In my environment, it solves the problem. thanks.

PR #1439, opened 2021-09-09
[bugfix] fix unknown status problems in search operation

kpango:

### Description:
In this PR, I have fixed the problem of unexpected Unknown state when searching in Vald.
There were multiple problems that occurred in unison with this problem.

1. The Agent was returning a nil response or an Internal error response instead of an Empty response when returning zero search results (cases where there was no approximate vector in radius or during CreateIndex).
2. The pkg agent only checks for error status during CreateIndex, so it now correctly handles EmptySearch errors as well
3. The Code in internal/core/algorithm/ngt.Search does not decide whether a response is Empty or not, and only handles CGO Error as an error, so it is necessary to return Empty Error for Empty Response.
4. Merge Error in internal/errgroup had a problem. The error was wrapped as a new string error without keeping the type information of the first goroutine thread, and the Error attribute could not be determined.
5. When the Gateway Aggregates the search results, it does a lot of comparison of floating point numbers, but in the Go language, the comparison is not accurate for very small numbers of floating points in the normal implementation, and zero matches are likely to occur due to missing information of small Distance. so I changed implementation to using math/big package.

1. Perform a search immediately after starting the Agent alone and confirm that NotFound is returned.
2. Confirmed that ErrCreateIndexIsInprogress or ErrEmptySearchResult is returned for a single agent after inserting multiple items and executing CreateIndex at the same time as the search.
3. Confirmed that "Unknown" is not returned and "NotFound" is returned successfully by searching Vald Cluster without Insert.

### Environment:

- Go Version: 1.17
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 1.12.3

### Types of changes:

- [x] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/core/algorithm/ngt/ngt.go
+++ b/internal/core/algorithm/ngt/ngt.go
@@ -275,7 +275,7 @@ func (n *ngt) create() (err error) {
 
 func (n *ngt) open() error {
 	if !file.Exists(n.idxPath) {
-		return errors.ErrIndexNotFound
+		return errors.ErrIndexFileNotFound
 	}
 
 	path := C.CString(n.idxPath)
--- a/internal/core/algorithm/ngt/ngt.go
+++ b/internal/core/algorithm/ngt/ngt.go
@@ -353,7 +353,7 @@ func (n *ngt) Search(vec []float32, size int, epsilon, radius float32) (result [
 		if err != nil {
 			err = errors.ErrEmptySearchResult
 		}
-		return err
+		return nil, err
 	}
 	// switch rsize := int(C.ngt_get_result_size(results, n.ebuf)); rsize {
 	// case -1:
--- a/internal/core/algorithm/ngt/ngt.go
+++ b/internal/core/algorithm/ngt/ngt.go
@@ -305,7 +305,7 @@ func (n *ngt) loadObjectSpace() error {
 }
 
 // Search returns search result as []SearchResult.
-func (n *ngt) Search(vec []float32, size int, epsilon, radius float32) ([]SearchResult, error) {
+func (n *ngt) Search(vec []float32, size int, epsilon, radius float32) (result []SearchResult, err error) {
 	if len(vec) != int(n.dimension) {
 		return nil, errors.ErrIncompatibleDimensionSize(len(vec), int(n.dimension))
 	}
@@ -345,16 +345,18 @@ func (n *ngt) Search(vec []float32, size int, epsilon, radius float32) ([]Search
 		n.mu.RUnlock()
 		return nil, n.newGoError(ne)
 	}
-
 	n.mu.RUnlock()
 
-	rsize := int(C.ngt_get_result_size(results, n.ebuf))
-	if rsize == -1 {
+	switch rsize := int(C.ngt_get_result_size(results, n.ebuf)); rsize {
+	case -1:
 		return nil, n.newGoError(n.ebuf)
+	case 0:
+		return nil, errors.ErrEmptySearchResult
+	default:
+		result = make([]SearchResult, rsize)
 	}
 
-	result := make([]SearchResult, rsize)
-	for i := 0; i < rsize; i++ {
+	for i := range result {
 		d := C.ngt_get_result(results, C.uint32_t(i), n.ebuf)
 		if d.id == 0 && d.distance == 0 {
 			result[i] = SearchResult{0, 0, n.newGoError(n.ebuf)}
--- a/internal/core/algorithm/ngt/ngt.go
+++ b/internal/core/algorithm/ngt/ngt.go
@@ -347,14 +347,23 @@ func (n *ngt) Search(vec []float32, size int, epsilon, radius float32) (result [
 	}
 	n.mu.RUnlock()
 
-	switch rsize := int(C.ngt_get_result_size(results, n.ebuf)); rsize {
-	case -1:
-		return nil, n.newGoError(n.ebuf)
-	case 0:
-		return nil, errors.ErrEmptySearchResult
-	default:
-		result = make([]SearchResult, rsize)
+	rsize := int(C.ngt_get_result_size(results, n.ebuf))
+	if rsize <= 0 {
+		err = n.newGoError(n.ebuf)
+		if err != nil {
+			err = errors.ErrEmptySearchResult
+		}
+		return err
 	}
+	// switch rsize := int(C.ngt_get_result_size(results, n.ebuf)); rsize {
+	// case -1:
+	// 	return nil, n.newGoError(n.ebuf)
+	// case 0:
+	// 	return nil, errors.ErrEmptySearchResult
+	// default:
+	// 	result = make([]SearchResult, rsize)
+	// }
+	result = make([]SearchResult, rsize)
 
 	for i := range result {
 		d := C.ngt_get_result(results, C.uint32_t(i), n.ebuf)
--- a/internal/core/algorithm/ngt/ngt.go
+++ b/internal/core/algorithm/ngt/ngt.go
@@ -350,7 +350,7 @@ func (n *ngt) Search(vec []float32, size int, epsilon, radius float32) (result [
 	rsize := int(C.ngt_get_result_size(results, n.ebuf))
 	if rsize <= 0 {
 		err = n.newGoError(n.ebuf)
-		if err != nil {
+		if err == nil {
 			err = errors.ErrEmptySearchResult
 		}
 		return nil, err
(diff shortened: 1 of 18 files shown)


internal/core/algorithm/ngt/ngt.go
```diff
@@ -345,16 +345,18 @@ func (n *ngt) Search(vec []float32, size int, epsilon, radius float32) ([]Search
...
-
 	n.mu.RUnlock()
 
-	rsize := int(C.ngt_get_result_size(results, n.ebuf))
-	if rsize == -1 {
+	switch rsize := int(C.ngt_get_result_size(results, n.ebuf)); rsize {
+	case -1:
 		return nil, n.newGoError(n.ebuf)
+	case 0:
+		return nil, errors.ErrEmptySearchResult
```
rinx, line 354: Please add a new test case that it correctly returns `ErrEmptySearchResult` when there's no index.

pkg/agent/core/ngt/service/ngt_stateful_test.go
```diff
@@ -515,8 +516,7 @@ var (
 			result commands.Result,
 		) *gopter.PropResult {
 			rc := result.(*resultContainer)
-
-			if rc.err != nil {
+			if rc.err != nil && !errors.Is(rc.err, errors.ErrEmptySearchResult) {
```
rinx, line 519: Please check `rc.err` is `ErrEmptySearchResult` when all of the idA, idB, idC are not indexed.

pkg/agent/core/ngt/service/ngt_stateful_test.go
```diff
@@ -562,7 +562,7 @@ var (
 		) *gopter.PropResult {
 			rc := result.(*resultContainer)
 
-			if rc.err != nil {
+			if rc.err != nil && !errors.Is(rc.err, errors.ErrEmptySearchResult) {
```
rinx, line 565: same.

pkg/agent/core/ngt/service/ngt_stateful_test.go
```diff
@@ -608,7 +608,7 @@ var (
...
```
rinx, line 611: same.

pkg/agent/core/ngt/service/ngt_stateful_test.go
```diff
@@ -654,7 +654,7 @@ var (
...
```
rinx, line 657: same.

rinx (changes requested):

pkg/agent/core/ngt/service/ngt_stateful_test.go
rinx: oops, `searchByID[ABC]Command` doesn't be fired when there's no index.
Please remove `&& !errors.Is(rc.err, errors.ErrEmptySearchResult)` from these lines.

kpango: It seems that the search is being performed even if there is no Index, and if I delete them, PBT fails with ErrEmptySearchResult every time.

kpango: anyway I'll remove them.

kpango: As you can see, the searchbyid operation has failed.
It looks like we need to check the State for the same as Search.
```
! NGT: Falsified after 1 passed tests.
> Labels of failing property: SearchByID-B
error
search result is empty
ARG_0: initialState=&{map[id-a:1 id-b:1 id-c:1] map[id-a:0 id-b:0 id-c:0]}
   sequential=[Insert-B CreateIndex SearchByID-B]
ARG_0_ORIGINAL (5 shrinks): initialState=&{map[id-a:1 id-b:1 id-c:1]
   map[id-a:0 id-b:0 id-c:0]} sequential=[Exists-A Insert-B Update-B
   Exists-B Insert-A Exists-A Exists-B CreateIndex SearchByID-B Exists-A]
Elapsed time: 31.366966318s
--- FAIL: TestStatefulNGT (31.37s)
    properties.go:57: failed with initial seed: 1631243130895981307
FAIL
```

pkg/agent/core/ngt/service/ngt_stateful_test.go
```diff
@@ -514,9 +515,13 @@ var (
...
 		) *gopter.PropResult {
+			st := state.(*ngtState)
 			rc := result.(*resultContainer)
-
-			if rc.err != nil {
+			if rc.err != nil &&
+				!(st.states[idA] != INDEXED &&
+					st.states[idB] != INDEXED &&
+					st.states[idC] != INDEXED &&
+					errors.Is(rc.err, errors.ErrEmptySearchResult)) {
```
rinx, line 524: I think it should be like the following.
```go
			if rc.err != nil {
				if st.states[idA] != INDEXED &&
					st.states[idB] != INDEXED &&
					st.states[idC] != INDEXED {
					if !errors.Is(rc.err, errors.ErrEmptySearchResult) {
						return &gopter.PropResult{
							Status: gopter.PropFalse,
							Error:  rc.err,
							Labels: []string{
								"Search",
								"there's no index but it doesn't return ErrEmptySearchResult",
								rc.err.Error(),
							},
						}
					}
				} else {
					return &gopter.PropResult{
						Status: gopter.PropFalse,
						Error:  rc.err,
						Labels: []string{
							"Search",
							"error",
							rc.err.Error(),
						},
					}
				}
			}
```
It is easier to find what is problem.

internal/net/grpc/status/status_test.go
```diff
@@ -1714,7 +1715,41 @@ func TestParseError(t *testing.T) {
 		return nil
 	}
 	tests := []test{
-		// TODO test cases
+		// func() test {
```
rinx, line 1718: Why is this test code commented out?

pkg/agent/core/ngt/service/ngt_stateful_test.go
rinx: I think it indicates there's some bugs in the implementation code.

> sequential=[Insert-B CreateIndex SearchByID-B]

This operation sequence seems to be correct, right?

kpango: I don't know the cause of this decrease in PBT, so I left this as a TODO comment to be investigated at a later date.
This PR is a bug fix and needs to be released as soon as possible, so I'll work around it for now.

internal/net/grpc/status/status_test.go
kpango: I wrote the test code for the verification of this bug, and I thought it would be useful for the SET team to start working on the test code task in the future, so I left a Sample test case.
I'll leave a comment on this.

pkg/agent/core/ngt/service/ngt_stateful_test.go
```diff
@@ -561,13 +577,31 @@ var (
...
+						}
+					}
+					return &gopter.PropResult{Status: gopter.PropTrue}
+				}
 				return &gopter.PropResult{
 					Status: gopter.PropFalse,
 					Error:  rc.err,
 					Labels: []string{
-						"SearchByID-A",
+						"Search",
```
rinx, line 604: You should not rename it.

pkg/agent/core/ngt/service/ngt_stateful_test.go
```diff
@@ -607,13 +641,31 @@ var (
...
+						}
+					}
+					return &gopter.PropResult{Status: gopter.PropTrue}
+				}
 				return &gopter.PropResult{
 					Status: gopter.PropFalse,
 					Error:  rc.err,
 					Labels: []string{
-						"SearchByID-B",
+						"Search",
```
rinx, line 668: same.

pkg/agent/core/ngt/service/ngt_stateful_test.go
```diff
@@ -653,13 +705,31 @@ var (
...
+						}
+					}
+					return &gopter.PropResult{Status: gopter.PropTrue}
+				}
 				return &gopter.PropResult{
 					Status: gopter.PropFalse,
 					Error:  rc.err,
 					Labels: []string{
-						"SearchByID-C",
+						"Search",
```
rinx, line 732: same here.

rinx (changes requested):

internal/net/grpc/status/status_test.go
rinx: I already know about it.

If the test case works correctly, you don't need to comment it out. I asked like that.

PR #1530, opened 2022-01-06
Implement test for net/grpc codec,logger and server

kevindiu:
Signed-off-by: kevindiu <kevindiujp@gmail.com>

### Description:

This PR implements the following tests for net/grpc package.
- codec.go
- logger/*
- server.go

### Environment:

- Go Version: 1.17.5
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 1.13.8

### Types of changes:

- [ ] Bug fix [type/bug]
- [x] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/internal/net/grpc/logger/logger.go
+++ b/internal/net/grpc/logger/logger.go
@@ -30,6 +30,11 @@ type logger struct {
 
 var once sync.Once
 
+const (
+	tag = "[gRPC Log]"
+)
+
+// Init initialize the logging level and the logger.
 func Init() {
 	once.Do(func() {
 		var v int
@@ -40,58 +45,67 @@ func Init() {
 	})
 }
 
-const (
-	tag = "[gRPC Log]"
-)
-
+// Info prints the debug log to the logger.
 func (l *logger) Info(args ...interface{}) {
 	log.Debug(append([]interface{}{tag}, args...)...)
 }
 
+// Infoln prints the debug log to the logger.
 func (l *logger) Infoln(args ...interface{}) {
 	log.Debug(append([]interface{}{tag}, args...)...)
 }
 
+// Infof prints the debug log to the logger.
 func (l *logger) Infof(format string, args ...interface{}) {
 	log.Debugf(tag+"\t"+format, args...)
 }
 
+// Warning prints the warning log to the logger.
 func (l *logger) Warning(args ...interface{}) {
 	log.Warn(append([]interface{}{tag}, args...)...)
 }
 
+// Warningln prints the warning log to the logger.
 func (l *logger) Warningln(args ...interface{}) {
 	log.Warn(append([]interface{}{tag}, args...)...)
 }
 
+// Warningf prints the warning log to the logger.
 func (l *logger) Warningf(format string, args ...interface{}) {
 	log.Warnf(tag+"\t"+format, args...)
 }
 
+// Error prints the error log to the logger.
 func (l *logger) Error(args ...interface{}) {
 	log.Error(append([]interface{}{tag}, args...)...)
 }
 
+// Errorln prints the error log to the logger.
 func (l *logger) Errorln(args ...interface{}) {
 	log.Error(append([]interface{}{tag}, args...)...)
 }
 
+// Errorf prints the error log to the logger.
 func (l *logger) Errorf(format string, args ...interface{}) {
 	log.Errorf(tag+"\t"+format, args...)
 }
 
+// Fatal prints the fatal log to the logger and exit the program.
 func (l *logger) Fatal(args ...interface{}) {
 	log.Fatal(append([]interface{}{tag}, args...)...)
 }
 
+// Fatalln prints the fatal log to the logger and exit the program.
 func (l *logger) Fatalln(args ...interface{}) {
 	log.Fatal(append([]interface{}{tag}, args...)...)
 }
 
+// Fatalf prints the fatal log to the logger and exit the program.
 func (l *logger) Fatalf(format string, args ...interface{}) {
 	log.Fatalf(tag+"\t"+format, args...)
 }
 
+// V returns if the v is less than the verbosity level.
 func (l *logger) V(v int) bool {
 	return v <= l.v
 }
--- a/internal/net/grpc/logger/logger_test.go
+++ b/internal/net/grpc/logger/logger_test.go
@@ -28,21 +28,19 @@ import (
 )
 
 func TestInit(t *testing.T) {
-	type want struct{}
 	type test struct {
 		name       string
-		want       want
-		checkFunc  func(want) error
+		checkFunc  func() error
 		beforeFunc func(*testing.T)
 		afterFunc  func()
 	}
-	defaultCheckFunc := func(w want) error {
+	defaultCheckFunc := func() error {
 		return nil
 	}
 	tests := []test{
 		{
 			name: "set logger success with verbosity level is not set",
-			checkFunc: func(w want) error {
+			checkFunc: func() error {
 				if grpclog.V(1) {
 					return errors.New("verbosity level is set")
 				}
@@ -57,7 +55,7 @@ func TestInit(t *testing.T) {
 			beforeFunc: func(t *testing.T) {
 				t.Setenv("GRPC_GO_LOG_VERBOSITY_LEVEL", "2")
 			},
-			checkFunc: func(w want) error {
+			checkFunc: func() error {
 				if !grpclog.V(1) {
 					return errors.New("verbosity level 1 is not set")
 				}
@@ -90,7 +88,7 @@ func TestInit(t *testing.T) {
 			}
 
 			Init()
-			if err := test.checkFunc(test.want); err != nil {
+			if err := test.checkFunc(); err != nil {
 				tt.Errorf("error = %v", err)
 			}
 		})
--- a/internal/net/grpc/logger/logger_test.go
+++ b/internal/net/grpc/logger/logger_test.go
@@ -110,7 +110,7 @@ func Test_logger_Info(t *testing.T) {
 	}
 	tests := []test{
 		{
-			name: "info log the message",
+			name: "info success to log the message",
 			args: args{
 				args: []interface{}{"log message"},
 			},
--- a/internal/net/grpc/logger/logger_test.go
+++ b/internal/net/grpc/logger/logger_test.go
@@ -168,7 +168,7 @@ func Test_logger_Infoln(t *testing.T) {
 	}
 	tests := []test{
 		{
-			name: "infoln log the message",
+			name: "infoln success to log the message",
 			args: args{
 				args: []interface{}{"log message"},
 			},
@@ -227,7 +227,7 @@ func Test_logger_Infof(t *testing.T) {
 	}
 	tests := []test{
 		{
-			name: "infof log the message",
+			name: "infof success to log the message",
 			args: args{
 				args: []interface{}{"log message"},
 			},
@@ -285,7 +285,7 @@ func Test_logger_Warning(t *testing.T) {
 	}
 	tests := []test{
 		{
-			name: "warning log the message",
+			name: "warning success to log the message",
 			args: args{
 				args: []interface{}{"log message"},
 			},
@@ -343,7 +343,7 @@ func Test_logger_Warningln(t *testing.T) {
 	}
 	tests := []test{
 		{
-			name: "warningln log the message",
+			name: "warningln success to log the message",
 			args: args{
 				args: []interface{}{"log message"},
 			},
@@ -402,7 +402,7 @@ func Test_logger_Warningf(t *testing.T) {
 	}
 	tests := []test{
 		{
-			name: "warningf log the message",
+			name: "warningf success to log the message",
 			args: args{
 				args: []interface{}{"log message"},
 			},
@@ -460,7 +460,7 @@ func Test_logger_Error(t *testing.T) {
 	}
 	tests := []test{
 		{
-			name: "error log the message",
+			name: "error success to log the message",
 			args: args{
 				args: []interface{}{"log message"},
 			},
@@ -518,7 +518,7 @@ func Test_logger_Errorln(t *testing.T) {
 	}
 	tests := []test{
 		{
-			name: "errorln log the message",
+			name: "errorln success to log the message",
 			args: args{
 				args: []interface{}{"log message"},
 			},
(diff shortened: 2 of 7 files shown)


internal/net/grpc/logger/logger_test.go
```diff
@@ -29,41 +30,48 @@ func TestInit(t *testing.T) {
...
+		// 		}
+		// 		if !grpclog.V(2) {
+		// 			//	return errors.New("verbosity level is not set")
+		// 		}
+		// 		if grpclog.V(3) {
+		// 			return errors.New("verbosity level is not correct")
+		// 		}
+		// 		return nil
+		// 	},
+		// },
```
vankichi, line 66: WIP?

internal/net/grpc/logger/logger_test.go
```diff
@@ -101,37 +109,15 @@ func Test_logger_Info(t *testing.T) {
...
-		           fields: fields {
-		           v: 0,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "info log the message",
```
vankichi, line 113: please fix according to the guideline :)

internal/net/grpc/logger/logger_test.go
```diff
@@ -181,37 +167,15 @@ func Test_logger_Infoln(t *testing.T) {
...
-		           fields: fields {
-		           v: 0,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "infoln log the message",
```
vankichi, line 171: please fix it according to the guideline :)

internal/net/grpc/logger/logger_test.go
```diff
@@ -262,39 +226,15 @@ func Test_logger_Infof(t *testing.T) {
...
-		           fields: fields {
-		           v: 0,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "infof log the message",
```
vankichi, line 230: please fix it according to the guideline :)

internal/net/grpc/logger/logger_test.go
```diff
@@ -344,37 +284,15 @@ func Test_logger_Warning(t *testing.T) {
...
-		           fields: fields {
-		           v: 0,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "warning log the message",
```
vankichi, line 288: please fix it according to the guideline :)

internal/net/grpc/logger/logger_test.go
```diff
@@ -424,37 +342,15 @@ func Test_logger_Warningln(t *testing.T) {
...
-		           fields: fields {
-		           v: 0,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "warningln log the message",
```
vankichi, line 346: please fix it according to the guideline :)

internal/net/grpc/logger/logger_test.go
```diff
@@ -505,39 +401,15 @@ func Test_logger_Warningf(t *testing.T) {
...
-		           fields: fields {
-		           v: 0,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "warningf log the message",
```
vankichi, line 405: please fix it according to the guideline :)

internal/net/grpc/logger/logger_test.go
```diff
@@ -587,37 +459,15 @@ func Test_logger_Error(t *testing.T) {
...
-		           fields: fields {
-		           v: 0,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "error log the message",
```
vankichi, line 463: please fix it according to the guideline :)

internal/net/grpc/logger/logger_test.go
```diff
@@ -667,37 +517,15 @@ func Test_logger_Errorln(t *testing.T) {
...
-		           fields: fields {
-		           v: 0,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "errorln log the message",
```
vankichi, line 521: please fix it according to the guideline :)

internal/net/grpc/logger/logger_test.go
```diff
@@ -748,39 +576,15 @@ func Test_logger_Errorf(t *testing.T) {
...
-		           fields: fields {
-		           v: 0,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "errorf log the message",
```
vankichi, line 580: please fix it according to the guideline :)

internal/net/grpc/logger/logger_test.go
```diff
@@ -830,37 +635,18 @@ func Test_logger_Fatal(t *testing.T) {
...
+		// 	args: args{
+		// 		args: []interface{}{"log message"},
+		// 	},
+		// 	fields: fields{
+		// 		v: 0,
+		// 	},
+		// 	afterFunc: func(a args) {
+		// 		_ = recover()
+		// 	},
+		// },
```
vankichi, line 649: WIP?

internal/net/grpc/logger/logger_test.go
```diff
@@ -910,37 +696,18 @@ func Test_logger_Fatalln(t *testing.T) {
...
```
vankichi, line 710: WIP?

internal/net/grpc/logger/logger_test.go
```diff
@@ -991,39 +758,18 @@ func Test_logger_Fatalf(t *testing.T) {
...
```
vankichi, line 772: WIP?

internal/net/grpc/logger/logger_test.go
kevindiu: this test is ignored due to `Init()`  is executed automatically before the test is run. 
Should I remove this comment or should I write some comment about it?

internal/net/grpc/logger/logger_test.go
kevindiu: fatal function uses `os.Exit` to exit the program after the fatal message is logged. so I comment out the test case.
I have written the reason on line 616

internal/net/grpc/logger/logger_test.go
kevindiu: same as above

internal/net/grpc/logger/logger_test.go
kevindiu: same as above

internal/net/grpc/logger/logger_test.go
kevindiu:
```suggestion
			name: "info success to log the message",
```

internal/net/grpc/logger/logger_test.go
kmrmt: For reference, I found the old way to test with `os.Exit`, but I don't know it is suitable.
https://talks.golang.org/2014/testing.slide#23

internal/net/grpc/logger/logger_test.go
vankichi: umm, we should consider how to run the `set log level` case.

internal/net/grpc/logger/logger_test.go
vankichi: we can try it.

vankichi:
could you please check the comments from golangci lint and consider to fix?

internal/net/grpc/logger/logger_test.go
kevindiu: how about avoiding using inner function in `init()` and create a separate function to make it testable?

vankichi: can we separate the test functions?
e.g. `TestInit` -> test with no set value, `TestInit_SetEnv` -> test with setting env value.

if we could like it, we do not have to change the current implementation.

kevindiu: We can't do it. 
The reason is that the `init()` is called when we import the package, the `once` object is initialized when we import it.
So the call from the second time (by the test code) will not work as the `once` object is initialized and the inner function will not executed at the second time.

internal/net/grpc/logger/logger_test.go
kevindiu: I think it works, but there is a problem that is shows the test coverage is 0% on the `os.Exit` tests because of it create a sub-process to execute the test. 
The main process cannot read the coverage from the sub-process.
I will search for another method to test it.

kmrmt: I found the another way to test with os.Exit that enables to 100% coverage.
https://qiita.com/KEINOS/items/718c4237d8d0b3f4c26b#comment-2f55c169b593fb6733be
https://itnext.io/how-to-tdd-a-console-application-to-achieve-100-coverage-starting-from-main-test-go-934a617b080f

internal/net/grpc/logger/logger_test.go
vankichi: I see.
I wonder why you added this case, but we do not have to run.
If you'd like to add this case, could you please add any comments at the top of this case?

vankichi:
please resolve comments from golang ci lint
If the comment is a mistake, please click `ResoloveConversation` and add ignore comment if it works well.
Also, it's correct, please fix the codes.

internal/net/grpc/logger/logger_test.go
kevindiu:
```suggestion
		// this test case cannot be run because the once object is initialized
		// as the init() will be called when we import the package
		// {
		// 	name: "set logger success with verbosity level is set",
		// 	beforeFunc: func(t *testing.T) {
		// 		t.Setenv("GRPC_GO_LOG_VERBOSITY_LEVEL", "2")
		// 	},
		// 	checkFunc: func(w want) error {
		// 		if !grpclog.V(1) {
		// 			return errors.New("verbosity level 1 is not set")
		// 		}
		// 		if !grpclog.V(2) {
		// 			//	return errors.New("verbosity level is not set")
		// 		}
		// 		if grpclog.V(3) {
		// 			return errors.New("verbosity level is not correct")
		// 		}
		// 		return nil
		// 	},
		// },
```

internal/net/grpc/logger/logger_test.go
kevindiu: It doesn't work as the `os.exit` call is not from our side but from the 3rd party dependencies.
We cannot refactor the implementation from them so we cannot achieve 100% coverage in this case.

internal/net/grpc/logger/logger_test.go
```diff
@@ -29,41 +32,50 @@ func TestInit(t *testing.T) {
...
+		{
+			name: "set logger success with verbosity level is not set",
+			checkFunc: func(w want) error {
+				if grpclog.V(1) {
+					return errors.New("verbosity level is set")
+				}
+				return nil
+			},
+		},
+		// this test case cannot be run because the once object is initialized
```
hlts2, line 51: I think you should clear sync once object with before and after func

kevindiu: fixed

internal/net/grpc/logger/logger_test.go
```diff
@@ -29,41 +33,54 @@ func TestInit(t *testing.T) {
...
+				}
+				if grpclog.V(3) {
+					return errors.New("verbosity level is not correct")
+				}
+				return nil
+			},
+			afterFunc: func() {
+				once = sync.Once{}
+			},
+		},
```
vankichi, line 75: `want` is not set :thinking:

kevindiu: the expected behavior is checked in the checkfunc, and this function returns nothing, so it is not set. 🤔

vankichi: So, why you did not change `checkFunc`?

vankichi: it should be `checkFunc()` not `checkFunc(w want)`

kevindiu: fixed

vankichi (approved):

internal/net/grpc/logger/logger.go
```diff
@@ -40,58 +45,67 @@ func Init() {
...
 func (l *logger) Info(args ...interface{}) {
 	log.Debug(append([]interface{}{tag}, args...)...)
 }
 
+// Infoln prints the debug log to the logger.
 func (l *logger) Infoln(args ...interface{}) {
 	log.Debug(append([]interface{}{tag}, args...)...)
 }
 
+// Infof prints the debug log to the logger.
```
kmrmt, line 58: same as above

internal/net/grpc/logger/logger.go
```diff
@@ -40,58 +45,67 @@ func Init() {
...
-const (
-	tag = "[gRPC Log]"
-)
-
+// Info prints the debug log to the logger.
 func (l *logger) Info(args ...interface{}) {
 	log.Debug(append([]interface{}{tag}, args...)...)
 }
 
+// Infoln prints the debug log to the logger.
```
kmrmt, line 53: same as above

internal/net/grpc/logger/logger.go
```diff
@@ -40,58 +45,67 @@ func Init() {
 	})
 }
 
-const (
-	tag = "[gRPC Log]"
-)
-
+// Info prints the debug log to the logger.
```
kmrmt, line 48: Does Info print debug log?

internal/net/grpc/logger/logger_test.go
```diff
@@ -87,51 +102,27 @@ func Test_logger_Info(t *testing.T) {
...
-		           fields: fields {
-		           v: 0,
-		           },
-		           want: want{},
-		           checkFunc: defaultCheckFunc,
-		       }
-		   }(),
-		*/
+		{
+			name: "info success to log the message",
```
kmrmt, line 118: I think the function name "info" should start with a capital case "Info".

kevindiu: fixed

vankichi (approved):

kmrmt (approved):

vankichi (approved):

kmrmt (approved):

PR #1557, opened 2022-02-16
implement agent ngt handler search test

kmrmt:
Signed-off-by: Kosuke Morimoto <ksk@vdaas.org>

### Description:

I listed up candidate test cases.

In almost cases, we must implement both of float32 and uint8, so I wrote the below representation.
- Testing type
  - float32 / uint8 / common
    - case n

Search_Config has some parameters but I use only Num parameter because Radius and Epsilon are very complex parameter in NGT algorithm.

### Environment:

- Go Version: 1.17.7
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 1.14.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [X] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/pkg/agent/core/ngt/handler/grpc/handler_test.go
+++ b/pkg/agent/core/ngt/handler/grpc/handler_test.go
@@ -444,9 +444,9 @@ func Test_server_Search(t *testing.T) {
 	defer cancel()
 
 	type args struct {
-		ctx context.Context
-		num int
-		req *payload.Search_Request
+		ctx       context.Context
+		insertNum int
+		req       *payload.Search_Request
 	}
 	type fields struct {
 		gen func(int, int) [][]float32
@@ -493,8 +493,8 @@ func Test_server_Search(t *testing.T) {
 			return nil, err
 		}
 
-		reqs := make([]*payload.Insert_Request, a.num)
-		for i, v := range f.gen(a.num, f.ngtCfg.Dimension) {
+		reqs := make([]*payload.Insert_Request, a.insertNum)
+		for i, v := range f.gen(a.insertNum, f.ngtCfg.Dimension) {
 			reqs[i] = &payload.Insert_Request{
 				Vector: &payload.Object_Vector{
 					Id:     strconv.Itoa(i),
@@ -615,8 +615,8 @@ func Test_server_Search(t *testing.T) {
 		{
 			name: "Equivalence Class Testing case 1.1: success search vector (type: uint8)",
 			args: args{
-				ctx: ctx,
-				num: 1000,
+				ctx:       ctx,
+				insertNum: 1000,
 				req: &payload.Search_Request{
 					Vector: convertVectorUint8ToFloat32(vector.GaussianDistributedUint8VectorGenerator(1, defaultDimensionSize)[0]),
 					Config: defaultSearch_Config,
@@ -635,8 +635,8 @@ func Test_server_Search(t *testing.T) {
 		{
 			name: "Equivalence Class Testing case 1.2: success search vector (type: float32)",
 			args: args{
-				ctx: ctx,
-				num: 1000,
+				ctx:       ctx,
+				insertNum: 1000,
 				req: &payload.Search_Request{
 					Vector: vector.GaussianDistributedFloat32VectorGenerator(1, defaultDimensionSize)[0],
 					Config: defaultSearch_Config,
@@ -653,8 +653,8 @@ func Test_server_Search(t *testing.T) {
 		{
 			name: "Equivalence Class Testing case 2.1: fail search vector with different dimension (type: uint8)",
 			args: args{
-				ctx: ctx,
-				num: 1000,
+				ctx:       ctx,
+				insertNum: 1000,
 				req: &payload.Search_Request{
 					Vector: convertVectorUint8ToFloat32(vector.GaussianDistributedUint8VectorGenerator(1, defaultDimensionSize+1)[0]),
 					Config: defaultSearch_Config,
@@ -674,8 +674,8 @@ func Test_server_Search(t *testing.T) {
 		{
 			name: "Equivalence Class Testing case 2.2: fail search vector with different dimension (type: float32)",
 			args: args{
-				ctx: ctx,
-				num: 1000,
+				ctx:       ctx,
+				insertNum: 1000,
 				req: &payload.Search_Request{
 					Vector: vector.GaussianDistributedFloat32VectorGenerator(1, defaultDimensionSize+1)[0],
 					Config: defaultSearch_Config,
@@ -695,8 +695,8 @@ func Test_server_Search(t *testing.T) {
 		{
 			name: "Boundary Value Testing case 1.1: success search with 0 value (min value) vector (type: uint8)",
 			args: args{
-				ctx: ctx,
-				num: 1000,
+				ctx:       ctx,
+				insertNum: 1000,
 				req: &payload.Search_Request{
 					Vector: fill(float32(uint8(0))),
 					Config: defaultSearch_Config,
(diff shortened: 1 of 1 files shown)


pkg/agent/core/ngt/handler/grpc/handler_test.go
```diff
@@ -457,7 +457,123 @@ func Test_server_Search(t *testing.T) {
...
+			- uint8, float32 common
+		    | same vector       | false | false | false | true | true | true |
+			| inserted          |     5 |    10 |    20 |    5 |   10 |   20 |
+			| Search_Config.Num |    10 |    10 |    10 |   10 |   10 |   10 |
+				- case 1: inserted 5 different vector, Num is 10 success
+				- case 2: inserted 10 different vector, Num is 10 success
+				- case 3: inserted 20 different vector, Num is 10 success
+				- case 4: inserted 5 same vector, Num is 10 success?
+				- case 5: inserted 10 same vector, Num is 10 success?
+				- case 6: inserted 20 same vector, Num is 10 success?
```
kevindiu, line 488: I think there are some cases that the vector is too far from the search vector and cannot be searched, do you think we need to consider these case also?

pkg/agent/core/ngt/handler/grpc/handler_test.go
```diff
@@ -457,7 +457,123 @@ func Test_server_Search(t *testing.T) {
...
+		- Boundary Value Testing
+		  	- uint8, float32 common
+				- case 1: min value vector success
+				- case 2: max value vector success
+				- case 3: 0 length vector fail?
+				- case 4: max length vector fail?
+				- case 5: nil vector fail
+			- uint8
+			- float32
+				- case 6: 0 value vector success
```
kevindiu, line 476: I think uint8 also need 0 value test

kmrmt: I intentionally excepted uint8 0 value test because it can be from 0 to 255.

pkg/agent/core/ngt/handler/grpc/handler_test.go
kmrmt: If my memory is correct, in the initial phase of NGT each node has dense connections.
So we need not consider the cases.
However, in the pkg decision table test, I don't know which pattern is better for inserting a few vectors or many vectors like below.

```
- case 1: insert 1000 different vectors, Num is 10 success
- case 2: insert 10000 different vectors, Num is 10 success
- case 3: insert 100000 different vectors, Num is 10 success
- case 4: insert 1000 same vectors, Num is 10 success(?)
- case 5: insert 10000 same vectors, Num is 10 success(?)
- case 6: insert 100000 same vectors, Num is 10 success(?)
```

In these cases, it takes more time than above.

pkg/agent/core/ngt/handler/grpc/handler_test.go
kevindiu: so is it the same test case as the following test case?
```
		  	- uint8, float32 common
				- case 1: min value vector success
```

pkg/agent/core/ngt/handler/grpc/handler_test.go
kevindiu: umm... I think many vectors in a case is fine as it is a more realistic use case of NGT.
@kmrmt @vankichi  what do you think?

pkg/agent/core/ngt/handler/grpc/handler_test.go
kmrmt: Yes, that's right.

kevindiu: IC, thank you :)

pkg/agent/core/ngt/handler/grpc/handler_test.go
kmrmt: I think the number of vectors to insert should be as large as possible.

kmrmt:
@kevindiu @vankichi I have implemented the search handler test and I have a question and problem.
# question
- I skipped implementing Boundary Value Testing case 3.1 because this test is the same as Boundary Value Testing case 1.1. How do you think whether to implement or not, that they will have the same implementation but their meaning is different strictly?

# problem
- To enable `tt.Parallel()`, the test fails and the errors are different results for each run like below. I have no idea to solve this.
```
--- FAIL: Test_server_Search (64.10s)
    --- FAIL: Test_server_Search/Decision_Table_Testing_case_1.1:_success_with_Search_Config.Num=10_from_5_different_vectors_(type:_uint8) (0.01s)
        /go/src/github.com/vdaas/vald/pkg/agent/core/ngt/handler/grpc/handler_test.go:1278: error = Capi : ngt_create_graph_and_tree() : Error: /tmp/ngt-20220302-78076-13qr1qo/NGT-1.14.3/lib/NGT/ObjectRepository.h:58: NGT::ObjectSpace: Cannot open the specified file /tmp/ngt-1648444918986314000/obj.
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xa8 pc=0x462abf7]
```

```
--- FAIL: Test_server_Search (45.71s)
    --- FAIL: Test_server_Search/Decision_Table_Testing_case_5.1:_success_with_Search_Config.Num=10_from_10_same_vectors_(type:_uint8) (0.01s)
        /go/src/github.com/vdaas/vald/pkg/agent/core/ngt/handler/grpc/handler_test.go:1278: error = Capi : ngt_create_graph_and_tree() : Error: std::bad_alloc
```

```
--- FAIL: Test_server_Search (45.65s)
    --- FAIL: Test_server_Search/Equivalence_Class_Testing_case_1.1:_success_search_vector_(type:_uint8) (0.01s)
        /Users/kmorimot/.ghq/src/github.com/vdaas/vald/pkg/agent/core/ngt/handler/grpc/handler_test.go:1278: error = Capi : ngt_create_graph_and_tree() : Error: /tmp/ngt-20220302-78076-13qr1qo/NGT-1.14.3/lib/NGT/Common.h:1804: NGT::Common: Not open the specified stream yet.
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xa8 pc=0x462abf7]
```

kevindiu:
The possible solution of your problem is that to set the index path for each test.

Since the index path is initialized on the package, all test shared the same index path.
https://github.com/vdaas/vald/blob/master/internal/core/algorithm/ngt/option.go#L43

When it runs in parallel mode, the index file is accessed concurrently so the error is occurred.

kevindiu:
For your question, I think they are meaning differently so I think we can keep them both.
For the implementation, I think it is ok to write comment to indicate the test is duplicated (like currently).

vankichi:
> * Boundary Value Testing

IMO, you can rename Boundary Value Testing 1.1 or remove 1.1 instead of 3.1.
e.g., `success search with 0 value vector from 1000 vectors (type: uint8)` to `success search with 0 value (min value) vector from 1000 vectors (type: uint8)`
For this case, 0 and the minimum value of unit8 are the same.

pkg/agent/core/ngt/handler/grpc/handler_test.go
```diff
@@ -436,90 +439,829 @@ func Test_server_Exists(t *testing.T) {
 
 func Test_server_Search(t *testing.T) {
 	t.Parallel()
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
 	type args struct {
 		ctx context.Context
+		num int
```
vankichi, line 448: I suppose it'd better to use `insertNum` instead of it.
If it is used for only insert parameter.

vankichi (approved):

kevindiu (approved):

kevindiu:
ci error occurred on internal test & pkg test, can you please help to investigate it?

PR #1575, opened 2022-03-07
add index manager component overview document

vankichi:

### Description:

SSIA

### Environment:

- Go Version: 1.17.7
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 1.14.1

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [x] Add documents [type/documentation]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:

### DIFF
--- a/docs/overview/component/index-manager.md
+++ b/docs/overview/component/index-manager.md
@@ -4,7 +4,7 @@ Vald Index Manager is the component that controls the indexing process of Vald A
 
 ## Responsibility
 
-Vald Index Manager has a unique and simple role in preventing stop-the-world.
+Vald Index Manager has a unique and simple role in controlling the indexing timing for all Vald Agent pods in the Vald cluster.
 
 It requires `Vald Discoverer` to fulfill its responsibility.
 
--- a/docs/overview/component/index-manager.md
+++ b/docs/overview/component/index-manager.md
@@ -22,7 +22,7 @@ Vald Index Manager uses these for the controlling indexing process.
 
 ### Controlling Indexing process
 
-When Vald Agent Pod creates or saves indexes on its container memory, it can not run the search process with the request query.
+When Vald Agent Pod creates or saves indexes on its container memory, it blocks all searching requests from the user and returns an error instead of a search result.
 
 Stop-the-world happens when all Vald Agent pods run the function involved in the indexing operation, e.g., `createIndex`, simultaneously.
 
@@ -34,7 +34,7 @@ The control process is Vald Index Manager sends `createIndex` requests for concu
 
 At the end of each process, Vald Index Manager updates the index information from each Vald Agent pod.
 
-Vald Index Manager runs this process repeatedly at set time intervals.
+Vald Index Manager runs this process periodically by set time intervals.
 
 <div class="notice">
 Concurrency means the number of Vald Agent pods for simultaneously sending requests for the indexing operation.
--- a/docs/overview/component/index-manager.md
+++ b/docs/overview/component/index-manager.md
@@ -1,0 +1,43 @@
+# Vald Index Manager
+
+Vald Index Manager is the component that controls the indexing process of Vald Agent Pods in the same Vald cluster.
+
+## Responsibility
+
+Vald Index Manager has a unique and simple role in preventing stop-the-world.
+
+It requires `Vald Discoverer` to fulfill its responsibility.
+
+## Features
+
+This chapter shows the main features to fulfill Vald Index Manager’s role.
+
+### Syncing data from Vald Discoverer
+
+Vald Index Manager gets the IP addresses of each Vald Agent pod from Vald Discoverer when its container starts.
+
+In addition, when IP address changes, Vald Index Manager gets the new IP address from Vald Discoverer.
+
+Vald Index Manager uses these for the controlling indexing process.
+
+### Controlling Indexing process
+
+When Vald Agent Pod creates or saves indexes on its container memory, it can not run the search process with the request query.
+
+Stop-the-world happens when all Vald Agent pods run the function involved in the indexing operation, e.g., `createIndex`, simultaneously.
+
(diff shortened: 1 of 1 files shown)


docs/overview/component/index-manager.md
```diff
@@ -0,0 +1,43 @@
+# Vald Index Manager
+
+Vald Index Manager is the component that controls the indexing process of Vald Agent Pods in the same Vald cluster.
+
+## Responsibility
+
+Vald Index Manager has a unique and simple role in preventing stop-the-world.
```
kevindiu, line 7: Is the role of index manager is to prevent stop-the-world?

It think the main responsibility is to control the timing of indexing, so to prevent the stop-the-world due to NGT's defect.

docs/overview/component/index-manager.md
```diff
@@ -0,0 +1,43 @@
...
+
+It requires `Vald Discoverer` to fulfill its responsibility.
+
+## Features
+
+This chapter shows the main features to fulfill Vald Index Manager’s role.
+
+### Syncing data from Vald Discoverer
```
kevindiu, line 15: I think it is not the main feature of index manager? 
I think it should be part of "control indexing process" section

docs/overview/component/index-manager.md
```diff
@@ -0,0 +1,43 @@
...
+
+Vald Index Manager gets the IP addresses of each Vald Agent pod from Vald Discoverer when its container starts.
+
+In addition, when IP address changes, Vald Index Manager gets the new IP address from Vald Discoverer.
+
+Vald Index Manager uses these for the controlling indexing process.
+
+### Controlling Indexing process
+
+When Vald Agent Pod creates or saves indexes on its container memory, it can not run the search process with the request query.
```
kevindiu, line 25:
```suggestion
When Vald Agent Pod creates or saves indexes on its container memory, it blocks all searching requests from the user and an error will be returned.
```

docs/overview/component/index-manager.md
```diff
@@ -0,0 +1,43 @@
...
+
+Vald Index Manager manages the indexing process of all Vald Agent pods to prevent this event.
+
+Vald Index Manager uses a Vald Agent pods' IP address list from Vald Discoverer and index information, including the stored index count, uncommitted index count, creating index count, and saving index count, from each Vald Agent pod.
+
+The control process is Vald Index Manager sends `createIndex` requests for concurrency simultaneously, sends a new request when the job is finished and continues until it sends to all agents.
+
+At the end of each process, Vald Index Manager updates the index information from each Vald Agent pod.
+
+Vald Index Manager runs this process repeatedly at set time intervals.
```
kevindiu, line 37:
```suggestion
Vald Index Manager runs this process periodically and when the number of uncommitted indexes exceeds the defined threshold.
```

docs/overview/component/index-manager.md
vankichi: yes, it's right.
I'll update.

docs/overview/component/index-manager.md
vankichi: it is also correct.
why I wrote such as is `control indexing process` section will be too long if I include it.
And more, syncing from the discoverer is not implemented in the `controlling indexing process`, so I split.

docs/overview/component/index-manager.md
vankichi:
```suggestion
When Vald Agent Pod creates or saves indexes on its container memory, it blocks all searching requests from the user and returns an error instead of a search result.
```

docs/overview/component/index-manager.md
vankichi: I suppose `when the number ~` is not correct :thinking: 
please see: https://github.com/vdaas/vald/blob/8377913a40ebc859fde5eb00e26a29fb8d19f9b7/pkg/manager/index/service/indexer.go#L197-L246

`execute` function does not check the uncommitted indexes with the threshold.
So, If I fix it, I'll update it like below.
```suggestion
Vald Index Manager runs this process periodically by set time intervals.
```

kevindiu: Yes, there are no uncommitted index threshold, sorry for my mistake.

`periodically` means that it execute by some intervals, I think we can omit the `set time intervals` part.
```suggestion
Vald Index Manager runs this process periodically.
```

docs/overview/component/index-manager.md
kevindiu: I understand why you are not including this part into the section, but I think it shouldn't be the main feature of Index manager as syncing data from discoverer does not provide any functionality to the user. 
What do you think?

docs/overview/component/index-manager.md
vankichi: well, we can.
It may be redundant, but we consider which is user-friendly because the user should control their time interval by setting their `valdrelease.yaml` even your suggestion is ok.
I just want your consideration.

docs/overview/component/index-manager.md
vankichi: IMO your feedback is true and false.
Certainly, this does not give the benefit directly.
it's difficult because if this function does not work well, the user cannot get the effect of the `Vald Index Manager`.
So, both are correct and we have to choose one.
I'm fine whichever, but I appreciate it if you give the some suggestions.

kevindiu: I think it also make sense. 
Considering this section is short, I think it should be fine without any changes on it.

docs/overview/component/index-manager.md
kevindiu: I think you are right.

docs/overview/component/index-manager.md
vankichi: thanks :)
I won't update anymore about it.

kevindiu (approved):

kmrmt (approved):

PR #1600, opened 2022-03-30
fix brand guidelines display missing

taisuou:
### Description:
The contents of p9 in brand guidelines are not visible .

Updated the pdf and the contents of p9 are visible.
スクリーンショット 2022-03-30 17 34


vankichi (approved):

vankichi:
There is no code change. So, I'd like to merge this PR even it does not get 2 or more approves.

PR #1617, opened 2022-04-12
refactor: improve memory high usage of vald-agent

kpango:

#### WHAT

This PR contains the following changes
- Change from `fmt.Sprintf` to `+` when generating the string
- Delete info.Get call in Exists method
- Improve error handling

#### WHY

`fmt.Sprintf` was using a large amount of memory when the upsert instruction was called at high frequency, which was discovered using pyroscope. It is not necessary to use `fmt.Sprintf` in all cases and should not be used if we do not format.
Also, inside the Exits method of agent, there is no need to add a stack trace.

### Environment:

- Go Version: 1.18
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 1.14.3

### Types of changes:

- [ ] Bug fix [type/bug]
- [ ] Add tests [type/test]
- [ ] Update CI [type/ci]

### Changes to Core Features:

### Checklist:


vankichi (dismissed):
LGTM

kmrmt (approved):

internal/info/info.go
```diff
@@ -247,7 +247,7 @@ func (d Detail) String() string {
 // Get returns parased Detail object.
 func (i info) Get() Detail {
 	i.prepare()
-	defaultURL := fmt.Sprintf("https://%s/tree/%s", valdRepo, i.detail.GitCommit)
+	defaultURL := "https://" + valdRepo + "/tree/" + i.detail.GitCommit
```
kevindiu, line 250: I think we can initialize it when we initialize `info`.

kevindiu: um... `defaultURL` in this function may not be used.
we can add default case to initialize url here https://github.com/vdaas/vald/blob/e4410016fa8201ec27f47ae5da10fea979298d52/internal/info/info.go#L263-L282

```
switch {
		case strings.HasPrefix(file, runtime.GOROOT()+"/src"):
		......
		case strings.Contains(file, "go/src/") && strings.Contains(file, valdRepo):
			url = strings.Replace(strings.SplitN(file, "go/src/", 2)[1]+"#L"+strconv.Itoa(line), valdRepo, "https://"+valdRepo+"/blob/"+i.detail.GitCommit, -1)
                 default:
                          url = "https://" + valdRepo + "/tree/" + i.detail.GitCommit
		}
```

hlts2: I don't think a string can be initialized when info is initialized.
For example, how about doing it in the `prepare` function?

kpango: how about we initialize it at prepare function and hold them at struct field ?
and runtime.GOROOT as well.

kpango: runtime.GOROOT() also have overhead to fetch environment variable.
FYI: https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/runtime/extern.go;l=243

hlts2: Thank you for your comment.
I have fixed: https://github.com/vdaas/vald/pull/1617/commits/12b8de8831532bebcf52f0e1abfb34379e139dd5

hlts2: @kpango 
Thank you for sharing. I will fix it soon!

hlts2: I have fixed based on two comment 👍 

https://github.com/vdaas/vald/pull/1617/commits/12b8de8831532bebcf52f0e1abfb34379e139dd5
https://github.com/vdaas/vald/pull/1617/commits/3b4ae955be33b8577886c067efedad5aae7d057b

internal/info/info.go
```diff
@@ -37,6 +37,8 @@ type Info interface {
 }
 
 type info struct {
+	baseURL  string // e.g https://github.com/vdaas/vald/tree/master
+	goRoot   string
```
kpango, line 41: goRoot should be in Detail struct and name should be GoRoot

internal/info/info.go
```diff
@@ -319,9 +320,11 @@ func (i *info) prepare() {
 		if len(i.detail.BuildCPUInfoFlags) == 0 && len(BuildCPUInfoFlags) != 0 {
 			i.detail.BuildCPUInfoFlags = strings.Split(strings.TrimSpace(BuildCPUInfoFlags), " ")
 		}
+		i.baseURL = "https://" + valdRepo + "/tree/" + i.detail.GitCommit
+		i.goRoot = runtime.GOROOT()
```
kpango, line 324: Each of them it's better to check length if length is not 0 we shouldn't overwrite variable.

kmrmt (approved):
LGTM

internal/info/info.go
kevindiu: `prepare()` is called every time when `Get()` is called, so it may not effective if we implement it in `prepare()`.
I think `goRoot` can be initialize when we start using this package.

kpango: can you please read code？

kevindiu (approved):

kmrmt (dismissed):

vankichi (dismissed):

kpango (commented):
LGTM

vankichi (approved):

vankichi (approved):

PR #1892, opened 2022-12-13
Refactor Agent, add canceled status for CreateIndex API and Update VQueue impl

kpango:

### Description:
This PR was created to correct several implementation issues that were found in the process of investigating bugs in the Vald Agent.
1. Since Vald Agent's Create Index API currently only returns "Internal" when an error occurs, an API Status was added to investigate whether the error was caused by a timeout. 
2. When Vald Agent's Create Index is called by multiple threads, the execution itself is executed one thread at a time, but multiple threads accumulate in the execution waiting loop, so we added an implementation to limit the waiting queue to one.
3. When many Insert Deletes are executed in multi-threading, the VQueue calculation order becomes extremely large and CreateIndex takes a lot of time, so the VQueue implementation was greatly revised.

Additional work:
1. added long test for ngt service
2. Added -short parameter to Makefile.d/test.mk to turn off the Long test by default.
3. made the timeout of tests executed in Makefile.d/test.mk variable, so that it can be overridden at runtime.
4. Added test assets to generate NegativeUniform vectors.

### Versions:

- Go Version: 1.19.4
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 1.14.8

### Checklist:


internal/test/data/vector/gen.go
```diff
@@ -82,6 +86,23 @@ func UniformDistributedFloat32VectorGenerator(n, dim int) [][]float32 {
 	return genF32Slice(n, dim, rand.Float32)
 }
 
+// NegativeUniformDistributedFloat32VectorGenerator returns n float32 vectors with dim dimension and their values under Uniform distribution
+func NegativeUniformDistributedFloat32VectorGenerator(n, dim int) (vecs [][]float32) {
+	left, right := dim/2, dim-dim/2
+	lvs := float32VectorGenerator(n, left, func() float32 {
```
kevindiu, line 92:
```suggestion
	lvs := genF32Slice(n, left, func() float32 {
```

vankichi (approved):

hlts2 (approved):
LGTM

PR #1939, opened 2023-02-02
[bugfix] add target length validation for each grpc client exection method

kpango:

### Description:
Vald's gRPC Client is implemented to remove Unhealthy Targets from the connection list and not query the Target in question until the new connection becomes Healthy.

In the event of network total loss, all targets become unhealthy, so Gateways that perform Broadcast and DoMulti instructions do not enter the processing loop and the error is not recognized, making it appear as if the client has succeeded.

As a solution, gRPC Client changed its policy to return the ErrClientConnectionNotFound error when the Target of gRPC Client is zero or the number of times the processing loop is executed is zero, judging that there is no Target.

### Versions:

- Go Version: 1.20
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 2.0.9

### Checklist:


hlts2 (approved):
LGTM

datelier (dismissed):

hlts2 (approved):
LGTM

hlts2 (approved):
LGTM

datelier (approved):

PR #1976, opened 2023-03-13
Added flg that can disable to ingress defaultBackend

takuyaymd:
SSIA (vald-lb-gateway, vald-filter-gateway)

### Description:

When running Vald with multiple Namespaces in the same Kubernetes cluster, I had to disable defaultBackend depending on the infrastructure situation, So I added a flag to disable defaultBackend.

### Versions:

- Go Version: 1.20.2
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 2.0.9

### Checklist:


charts/vald/values.yaml
```diff
@@ -1069,6 +1069,11 @@ gateway:
       # @schema {"name": "gateway.lb.ingress.servicePort", "type": "string"}
       # gateway.lb.ingress.servicePort -- service port to be exposed by ingress
       servicePort: grpc
+      # @schema {"name": "gateway.lb.ingress.defaultBackend", "type": "object"}
+      defaultBackend:
+        # @schema {"name": "gateway.lb.ingress.defaultBackend.enabled", "type": "boolean"}
+        # gateway.lb.ingress.defaultBackend.enabled -- gateway ingress defaultBackend enabled
+        enabled: true
```
vankichi, line 1076: It would be better to define the IngressBackend object.
Ref: https://kubernetes.io/docs/reference/kubernetes-api/service-resources/ingress-v1/#IngressBackend

e.g.
```suggestion
      # @schema {"name": "gateway.lb.ingress.defaultBackend", "type": "object"}
      defaultBackend:
        # @schema {"name": "gateway.lb.ingress.defaultBackend.enabled", "type": "boolean"}
        # gateway.lb.ingress.defaultBackend.enabled -- gateway ingress defaultBackend enabled
        enabled: true
        resource:
        ...
        service:
        ...
```

And also, it requires updating the helm template.

charts/vald/values.yaml
```diff
@@ -1058,7 +1058,7 @@ gateway:
       pathType: ImplementationSpecific
       # @schema {"name": "gateway.lb.ingress.enabled", "type": "boolean"}
       # gateway.lb.ingress.enabled -- gateway ingress enabled
-      enabled: false
+      enabled: true
```
hlts2, line 1061: It would be better to set `false`. 🙏  If we do not have a controller for ingress, we cannot create an ingress resource and this deployment will fail.
Since I think `values.yaml` should work under any environment, it makes sense to let the user set it to enable if we want to use ingress.

hlts2, line 1061:
```suggestion
      enabled: false
```
It would be better to set false. 🙏 If we do not have a controller for ingress, we cannot create an ingress resource and this deployment will fail.
Since I think values.yaml should work under any environment, it makes sense to let the user set it to enable if we want to use ingress.

charts/vald/values.yaml
takuyaymd: Thanks for the review.
Currently, Vald's defaultBackend.service is set to the same properties as the ingress backend.
https://github.com/vdaas/vald/blob/main/charts/vald/templates/gateway/filter/ing.yaml#L36-L37C1

If we were to allow fine-tuning of defaultBackend based on the Kubernets specification, there would be a difference between this behavior and the one described above.

If this is to be absorbed, a flag like `followIngressProperty` would need to be added to allow the defaultBackend to follow the ingress.backend as before, or to be configured on its own.

I'm thinking that it would be better to simply be able to control whether defaultBackend is enabled/disabled in this case.

charts/vald/values.yaml
takuyaymd: I am assuming that if `ingress:false`, this setting is ignored in the first place. (If I'm wrong, please point it out.)
https://github.com/vdaas/vald/blob/main/charts/vald/templates/gateway/filter/ing.yaml#L17

I am thinking that if this flag is set to false, defaultBackend will behave differently in the new version, which may confuse existing users.

charts/vald/values.yaml
vankichi: > Currently, Vald's defaultBackend.service is set to the same properties as the ingress backend.
> https://github.com/vdaas/vald/blob/main/charts/vald/templates/gateway/filter/ing.yaml#L36-L37C1

It's correct in the current.
IMO, it is the chance to refactor templates to apply both IngressBackend configurations for available.
It will make help users can select which option to use.
So, I think applying the IngressBackend configuration object in this situation would be better.

charts/vald/values.yaml
vankichi: @hlts2 said this change forces the existing users who don't use `ingress` to change their values.
also, all tutorials don't apply the `ingress` feature.
That's way @hlts2 suggested do not change the `gateway.lb.ingress.enabled` value.

hlts2: @vankichi Thanks for the additional information 🙏 
@takuyaymd I commented because there are cases where a user does not use Ingress, such as when running locally, so I thought it would be better to be able to enable the setting when needed 👍

charts/vald/templates/gateway/filter/ing.yaml
```diff
@@ -31,10 +31,12 @@ metadata:
     app.kubernetes.io/component: gateway-filter
   name: {{ $gateway.name }}-ingress
 spec:
+  {{- if and $gateway.ingress.defaultBackend.enabled }}
```
vankichi, line 34:
```suggestion
  {{- if $gateway.ingress.defaultBackend.enabled }}
```

charts/vald/templates/gateway/lb/ing.yaml
```diff
@@ -31,10 +31,12 @@ metadata:
     app.kubernetes.io/component: gateway-lb
   name: {{ $gateway.name }}-ingress
 spec:
+  {{- if and $gateway.ingress.defaultBackend.enabled }}
```
vankichi, line 34:
```suggestion
  {{- if $gateway.ingress.defaultBackend.enabled }}
```

vankichi (changes requested):

charts/vald/templates/gateway/filter/ing.yaml
takuyaymd: Sorry, I fixed it!

charts/vald/templates/gateway/lb/ing.yaml
takuyaymd: 🙇

kpango (approved):
LGTM

vankichi (approved):

PR #2022, opened 2023-05-01
add network policy

ykadowak:
### Description:

Added network policy with the configuration as in the table. Also only allowed traffic within the deployed namespace. Plus please note that the egress of the filter gateway is fully open due to the wide variety of user-defined filter configurations.

from \ to | agent | discoverer | filter gateway | lb gateway | index | User-defined Filters | k8s apiserver
-- | -- | -- | -- | -- | -- | -- | --
agent | N/A | ⛔ | ⛔ | ⛔ | ⛔ | ⛔ | ⛔
discoverer | ⛔ | N/A | ⛔ | ⛔ | ⛔ | ⛔ | ✅
filter gateway | ⛔ | ⛔ | N/A | ✅ | ⛔ | ✅ | ⛔
lb gateway | ✅ | ✅ | ⛔ | N/A | ⛔ | ⛔ | ⛔
index | ✅ | ✅ | ⛔ | ⛔ | N/A | ⛔ | ⛔
User-defined Filters | ⛔ | ⛔ | ⛔ | ⛔ | ⛔ | N/A | ⛔
k8s apiserver | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | N/A

Confirmed this configuration passes e2e with the steps below.

```bash
make k8s/vald/deploy HELM_VALUES=example/helm/values.yaml
make e2e
```

### Versions:

- Go Version: 1.20.3
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 2.0.9

### Checklist:

### Special notes for your reviewer:

This PR will be followed by these PRs below.
- [ ] e2e test on PR when there's helm template change
- [ ] User-defined network policy for things like OpenTelemetry

NetworkPolicy is off by default until these changes are done.


charts/vald/templates/common/deny-all.yaml
```diff
@@ -0,0 +1,28 @@
...
+#
+
+# Deny all the ingress/egress traffic
+{{- if .Values.defaults.networkPolicy.enabled }}
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: deny-all
+spec:
+  podSelector: {}
```
kpango, line 24: It would be better to select only Vald components and namespaces.

charts/vald/templates/gateway/filter/networkpolicy.yaml
```diff
@@ -0,0 +1,39 @@
...
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
+    - from:
+      - namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: kube-system
+  egress:
+    # allow all the egress to communicate with user-defined filters
```
kpango, line 37: I think we can exclude kube-system for filter request.

charts/vald/values.yaml
```diff
@@ -830,6 +830,13 @@ defaults:
       # @schema {"name": "defaults.observability.trace.enabled", "type": "boolean"}
       # defaults.observability.trace.enabled -- trace enabled
       enabled: false
+  # @schema {"name": "defaults.networkPolicy", "type": "object", "anchor": "networkPolicy"}
+  networkPolicy:
+    # @schema {"name": "defaults.networkPolicy.enabled", "type": "boolean"}
+    # defaults.networkPolicy.enabled -- if network policy enabled
+    # TODO: Change this to true after implementing user custom network policy parser
+    enabled: false
```
kpango, line 838: How can we modify this if a specific component wants to turn off NetworkPolicy?

charts/vald/templates/common/deny-all.yaml
ykadowak: Is this something we really want to do?

Setting it up that way will allow all the components traffic to and from the outside by default, which raises security concerns if there is a bad configuration, etc. Instead, I think it makes more sense to deny all the traffic and then only allowing traffic for components that really need external traffics, such as `filter gateway`.

charts/vald/templates/gateway/filter/networkpolicy.yaml
ykadowak: [1aea701](https://github.com/vdaas/vald/pull/2022/commits/1aea701ea176adde9abb091a3d13e2cc3653b933)

charts/vald/templates/common/deny-all.yaml
kpango: Sorry, I misunderstood.

However, I think this method of denying all network In/Out and then opening the destination target only for specific components may affect other components that are not managed by Vald, such as components deployed in the same Namespace.

It would be more flexible to control the appropriate In/Out for each component.

ykadowak: I get what you mean. Does this resolve your concern?
[5353240](https://github.com/vdaas/vald/pull/2022/commits/5353240aa0f777a0a0f0ca9f81d0222ca486918c)

charts/vald/values.yaml
ykadowak: I think it's not that easy to do such a modification.

While it is possible to implement such a feature, the implementation may become complex depending on the use case. For example, even if you open the egress of one component, it will not be meaningful if the corresponding ingress of the component is not open. Managing all of these ingress/egress relationships is not practical, I believe.

What kind of use case are you considering?

kpango: I fully understand the difficulty of implementing this.

However, testing to open or shutdown only some LB GW or Agents when deploying in ChaosNetwork or other special environments could be a potential usecase.

Please let me discuss whether or not to support this in this PR, since the difficulty and scope of implementation will be considerably expanded if implemented.

charts/vald/templates/common/deny-all.yaml
kpango: looks good, thank you.

charts/vald/values.yaml
ykadowak: We discussed offline and concluded that having an option of entirely enable/disable is enough for the use case. Thus no change is required for this PR.

ykadowak:
There is a know bug in `kube-linter` and caused the `K8s manifests / runner / kubelinter for vald chart` fail. The fix should be released soon so we are leaving it.
https://github.com/stackrox/kube-linter/pull/546

kpango (approved):
LGTM

hlts2 (approved):
LGTM!

PR #2034, opened 2023-05-22
[WIP] update devcontainer to run ngt locally

ykadowak:
### Versions:

- Go Version: 1.20.3
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 2.0.11

### Checklist:

- [ ] I have read the [CONTRIBUTING](https://github.com/vdaas/vald/blob/main/CONTRIBUTING.md) document and completed [our CLA agreement](https://cla-assistant.io/vdaas/vald).
- [ ] I have checked open [Pull Requests](https://github.com/vdaas/vald/pulls) for the similar feature or fixes?

PR #2040, opened 2023-05-24
refactor and add test for service/ngt.go

ykadowak:
### Description:

As a preparation for https://github.com/vdaas/vald/pull/2034, extracted the process of initializing the backup folder and added some tests.

### Versions:

- Go Version: 1.20.3
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 2.0.11

### Checklist:

### DIFF
--- a/pkg/agent/core/ngt/service/ngt.go
+++ b/pkg/agent/core/ngt/service/ngt.go
@@ -203,7 +203,7 @@ func (n *ngt) prepareFolders() (err error) {
 		sep := string(os.PathSeparator)
 		absPath, err := filepath.Abs(strings.ReplaceAll(n.path, sep+sep, sep))
 		if err != nil {
-			log.Warn("keep going with relative path: %w", err)
+			log.Warnf("keep going with relative path:\t%v", err)
 		} else {
 			n.path = absPath
 		}
--- a/pkg/agent/core/ngt/service/ngt.go
+++ b/pkg/agent/core/ngt/service/ngt.go
@@ -203,8 +203,8 @@ func (n *ngt) prepareFolders() (err error) {
 		sep := string(os.PathSeparator)
 		absPath, err := filepath.Abs(strings.ReplaceAll(n.path, sep+sep, sep))
 		if err != nil {
-			log.Warn(err)
-		} else if file.Exists(absPath) {
+			log.Warn("keep going with relative path: %w", err)
+		} else {
 			n.path = absPath
 		}
 		n.basePath = n.path
--- a/pkg/agent/core/ngt/service/ngt.go
+++ b/pkg/agent/core/ngt/service/ngt.go
@@ -159,30 +159,13 @@ func New(cfg *config.NGT, opts ...Option) (nn NGT, err error) {
 		}
 	}
 	if len(n.path) == 0 {
+		log.Info("index path setting is empty, starting vald agent with in-memory mode")
 		n.inMem = true
 	}
 
-	if n.enableCopyOnWrite && !n.inMem && len(n.path) != 0 {
-		sep := string(os.PathSeparator)
-		n.path, err = filepath.Abs(strings.ReplaceAll(n.path, sep+sep, sep))
-		if err != nil {
-			log.Warn(err)
-		}
-		n.basePath = n.path
-		n.oldPath = file.Join(n.basePath, oldIndexDirName)
-		n.path = file.Join(n.basePath, originIndexDirName)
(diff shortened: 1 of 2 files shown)


pkg/agent/core/ngt/service/ngt.go
```diff
@@ -215,6 +198,33 @@ func New(cfg *config.NGT, opts ...Option) (nn NGT, err error) {
 	return n, nil
 }
 
+func (n *ngt) prepareFolders() (err error) {
+	if n.enableCopyOnWrite && !n.inMem && len(n.path) != 0 {
+		sep := string(os.PathSeparator)
+		n.path, err = filepath.Abs(strings.ReplaceAll(n.path, sep+sep, sep))
+		if err != nil {
+			// TODO: maybe we should return error here?
+			log.Warn(err)
```
ykadowak, line 207: @kpango 
What do you think?

kpango: Our implementation before refactoring did not handle this error but just output it to the log and then continue processing.

There are 3 reasons for previous impl.
1. Abs fails in some cases such as os.Getwd (unix) and syscall.FullPath (windows), but it is rare for these to fail.
2. The motivation for using Abs Path is to eliminate uncertainty caused by relative paths.
3. Even if Abs Path fails, it is possible to start with relative path by using n.path.

However, the previous implementation directly rewrites n.path by filepath.Abs, so it should have been implemented as follows.
```
absPath, err := filepath.Abs(strings.ReplaceAll(n.path, sep+sep, sep))
if err != nil {
        log.Warn(err)
} else if absPath != "" {
        n.path = absPath
}
```

ykadowak: > There are 3 reasons for previous impl.

Understood. Thanks.

> However, the previous implementation directly rewrites n.path by filepath.Abs, so it should be implemented as follows.

I think there are no side effects, so I think it's fine, but are there any cases where `file.Exists(absPath)` actually becomes false after `filepath.Abs` successes?

ykadowak: any way implemented that way
[db561f2](https://github.com/vdaas/vald/pull/2040/commits/db561f21f9a32560ff43ba4645d030e90dbe757d)

kpango: > I think there are no side effects so it should be fine, but are there any cases where file.Exists(absPath) actually becomes false after filepath.Abs successes?
I checked go's internal code, there is no cases like that.
Can you please remove absPath checking with file.Exists?

pkg/agent/core/ngt/service/ngt.go
```diff
@@ -215,6 +198,34 @@ func New(cfg *config.NGT, opts ...Option) (nn NGT, err error) {
 	return n, nil
 }
 
+func (n *ngt) prepareFolders() (err error) {
+	if n.enableCopyOnWrite && !n.inMem && len(n.path) != 0 {
+		sep := string(os.PathSeparator)
+		absPath, err := filepath.Abs(strings.ReplaceAll(n.path, sep+sep, sep))
+		if err != nil {
+			log.Warn(err)
+		} else if file.Exists(absPath) {
```
kpango, line 207:
```suggestion
		} else {
```

kpango: sorry about that.

pkg/agent/core/ngt/service/ngt.go
ykadowak: 👍 
Also added some log msg if that's ok.
[5812658](https://github.com/vdaas/vald/pull/2040/commits/58126583e24d2c481340da257137c756cf192d91)

pkg/agent/core/ngt/service/ngt.go
```diff
@@ -215,6 +198,34 @@ func New(cfg *config.NGT, opts ...Option) (nn NGT, err error) {
 	return n, nil
 }
 
+func (n *ngt) prepareFolders() (err error) {
+	if n.enableCopyOnWrite && !n.inMem && len(n.path) != 0 {
+		sep := string(os.PathSeparator)
+		absPath, err := filepath.Abs(strings.ReplaceAll(n.path, sep+sep, sep))
+		if err != nil {
+			log.Warn("keep going with relative path: %w", err)
```
kpango, line 206:
```suggestion
			log.Warnf("keep going with relative path:\t%v", err)
```

vankichi (approved):

kevindiu (approved):

PR #2059, opened 2023-06-13
add decode kvsdb tool

datelier:
### Description:
add decode kvsdb tool.

run example:
```
go run main.go --format XXX --path XXX --file XXX --timestamp-file XXX
```

### Versions:

- Go Version: 1.20.3
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 2.0.11

### Checklist:

### DIFF
--- a/hack/tools/kvsdb/main.go
+++ b/hack/tools/kvsdb/main.go
@@ -1,0 +1,89 @@
+//
+// Copyright (C) 2019-2023 vdaas.org vald team <vald@vdaas.org>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package main
+
+import (
+	"context"
+	"encoding/gob"
+	"flag"
+	"fmt"
+	"io/fs"
+	"os"
+
+	"github.com/vdaas/vald/internal/file"
+	"github.com/vdaas/vald/internal/log"
+	"github.com/vdaas/vald/pkg/agent/core/ngt/service/kvs"
+)
+
+var (
+	format = flag.String("format", "csv", "file format(csv,tsv)")
+	kvsFileName = flag.String("file", "ngt-meta.kvsdb", "kvsdb file name")
+	kvsTimestampFileName = flag.String("timestamp-file", "ngt-timestamp.kvsdb", "kvsdb timestamp file name")
+	path = flag.String("path", ".", "kvsdb file path")
+)
+
+func main() {
+	flag.Parse()
+	log.Init()
+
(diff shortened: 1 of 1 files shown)


hack/tools/kvsdb/main.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// kvs load
+	for k, id := range m {
+		if ts, ok := mt[k]; ok {
+			kvsdb.Set(k, id, ts)
+		} else {
+			kvsdb.Set(k, id, 0)
+		}
+	}
+
```
kpango, line 77: it would be better to add header.

hack/tools/kvsdb/main.go
```diff
@@ -0,0 +1,91 @@
...
+	mt := make(map[string]int64)
+	gob.Register(map[string]int64{})
+	var ft *os.File
+	defer ft.Close()
+	ft, _ = file.Open(
+		file.Join(*path, *kvsTimestampFileName),
+		os.O_RDONLY|os.O_SYNC,
+		fs.ModePerm,
+	)
+	_ = gob.NewDecoder(ft).Decode(&mt)
```
kmrmt, line 65: I think you should stop this program like below when it returns errors because there can be the case such as no permission.
And I think you should consider logging levels.
This suggestion is just idea.
```suggestion
	var err error
	f, err = file.Open(
		file.Join(*path, *kvsFileName),
		os.O_RDONLY|os.O_SYNC,
		fs.ModePerm,
	)
	if err != nil {
		log.Fatal(err)
	}
	if err = gob.NewDecoder(f).Decode(&m); err != nil {
		log.Fatal(err)
	}

	// timestamp
	mt := make(map[string]int64)
	gob.Register(map[string]int64{})
	var ft *os.File
	defer ft.Close()
	ft, err = file.Open(
		file.Join(*path, *kvsTimestampFileName),
		os.O_RDONLY|os.O_SYNC,
		fs.ModePerm,
	)
	if err != nil {
		log.Fatal(err)
	}
	if err = gob.NewDecoder(ft).Decode(&mt); err != nil {
		log.Fatal(err)
	}
```

hack/tools/kvsdb/main.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	s = append(s, []string{"uuid", "oid", "timestamp"})
+	for k, id := range m {
+		if ts, ok := mt[k]; ok {
+			s = append(s, []string{k, strconv.FormatUint(uint64(id), 10), strconv.FormatInt(ts, 10)})
+		} else {
+			s = append(s, []string{k, strconv.FormatUint(uint64(id), 10), "0"})
+		}
+		if len(s)*int(unsafe.Sizeof("")) > 4e+6 {
+			w.WriteAll(s)
```
kmrmt, line 86: same as above
```suggestion
			err = w.WriteAll(s)
			if err != nil {
				log.Warn(err)
			}
```

hack/tools/kvsdb/main.go
```diff
@@ -0,0 +1,91 @@
...
+			s = nil
+		}
+	}
+	w.WriteAll(s)
```
kmrmt, line 90: same as above
```suggestion
	err = w.WriteAll(s)
	if err != nil {
		log.Fatal(err)
	}
```

kmrmt (changes requested):

kmrmt (approved):
LGTM

hlts2 (approved):
LGTM

PR #2091, opened 2023-06-27
Add backup origin when CoW enabled and failed to load primary

ykadowak:
### Description:

With this PR, the primary index will be backed up immediately when the agent fails to load it with CoW enabled. Before, it was backed up only when it fails to load both primary and old index.

### Versions:

- Go Version: 1.20.3
- Docker Version: 20.10.8
- Kubernetes Version: 1.22.0
- NGT Version: 2.0.11

### Checklist:

### DIFF
--- a/pkg/agent/core/ngt/service/ngt.go
+++ b/pkg/agent/core/ngt/service/ngt.go
@@ -655,7 +655,7 @@ func (n *ngt) initNGT(opts ...core.Option) (err error) {
 		} else {
 			log.Warnf("failed to load vald primary index from %s\t error: %v\ttrying to load from old copied index data from %s", n.path, err, n.oldPath)
 			if needsBackup(n.path) {
-				log.Infof("starting to backup broken index at %v", n.path)
+				log.Infof("starting to backup broken index at %s", n.path)
 				if err := n.backupBroken(ctx); err != nil {
 					log.Warnf("failed to backup broken index. will try to restart from old index anyway: %v", err)
 				}
--- a/pkg/agent/core/ngt/service/ngt.go
+++ b/pkg/agent/core/ngt/service/ngt.go
@@ -527,6 +527,11 @@ func (n *ngt) backupBroken(ctx context.Context) error {
 	atomic.SwapUint64(&n.nobic, uint64(len(files)))
 	log.Debugf("broken index count updated: %v", n.nobic)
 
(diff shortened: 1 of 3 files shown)


pkg/agent/core/ngt/service/ngt.go
```diff
@@ -655,6 +654,12 @@ func (n *ngt) initNGT(opts ...core.Option) (err error) {
 			)
 		} else {
 			log.Warnf("failed to load vald primary index from %s\t error: %v\ttrying to load from old copied index data from %s", n.path, err, n.oldPath)
+			if needsBackup(n.path) {
+				log.Infof("starting to backup broken index at %v", n.path)
```
kpango, line 658:
```suggestion
				log.Infof("starting to backup broken index at %s", n.path)
```

pkg/agent/core/ngt/service/ngt.go
```diff
@@ -527,6 +527,11 @@ func (n *ngt) backupBroken(ctx context.Context) error {
 	atomic.SwapUint64(&n.nobic, uint64(len(files)))
 	log.Debugf("broken index count updated: %v", n.nobic)
 
+	// remake the path since it has been moved to broken directory
+	if err := file.MkdirAll(n.path, fs.ModePerm); err != nil {
+		return errors.Join(errors.ErrIndexDirectoryRecreationFailed, err)
```
kpango, line 532: Since errors.Join adds backward errors to the first Error, so it is better to write new error definitions in the backward order.

```suggestion
		return errors.Join(err, errors.ErrIndexDirectoryRecreationFailed)
```


Also, the internal handling of ErrorSlice is not relevant in this case, but if err is already an ErrorSlice and err is the second argument, ErrorSlice allocations will occur because the first argument is an errString type. It is better to connect the first argument for the unknown Error type as much as possible.

kpango (changes requested):

pkg/agent/core/ngt/service/ngt.go
ykadowak: [e92ded8](https://github.com/vdaas/vald/pull/2091/commits/e92ded8d31c388949b7f28728a911f8a967a40d5)

kpango (approved):
LGTM

kmrmt (approved):

PR #2102, opened 2023-07-01
use GOBIN instead of GOPATH/bin

kpango:
### Description:
SSIA

### Versions:

- Go Version: 1.20.5
- Docker Version: 20.10.8
- Kubernetes Version: v1.27.3
- NGT Version: 2.0.13

### Checklist:


ykadowak:
It should be checked if `GOBIN` is set at first and if not fallback to `GOPATH/bin`?

kpango:
@ykadowak Indeed, I think it is a good idea to backport to GOPATH/bin when "go env GOBIN" is empty.
However, since GOPATH is also obtained from "go env GOPATH", I think it is a rare case that only "go env GOBIN" is missing in the go environment in the initial Install state.

vankichi (approved):

kevindiu (approved):

PR #2154, opened 2023-08-21
Refactor agent GetObject error not to wrap with details for performance issue

ykadowak:
### Description:

This PR removes error details from agent `GetObject` and `Exists` rpc error. This is required because the following performance issue was discovered while developing the `index correction` feature.

The issues is that when the client issues a large number of `GetObject` to the agent, the process of the http2 client huffman decoding the error detail from the header becomes a bottleneck in CPU processing, resulting in a decrease of QPS. Same goes to `Exists`.

Even if this error detail is removed, its impact is considered to be minor. This is because it only affects the use case of the agent standalone, which is limited. Also, even in that case, the client side can manage the error without details because it must be a unary request, not stream, to the agent.

### Versions:

- Go Version: 1.20.6
- Docker Version: 20.10.8
- Kubernetes Version: v1.27.3
- NGT Version: 2.0.16

### Checklist:

### DIFF
--- a/pkg/agent/core/ngt/handler/grpc/object.go
+++ b/pkg/agent/core/ngt/handler/grpc/object.go
@@ -21,7 +21,6 @@ import (
 	"github.com/vdaas/vald/apis/grpc/v1/payload"
 	"github.com/vdaas/vald/apis/grpc/v1/vald"
 	"github.com/vdaas/vald/internal/errors"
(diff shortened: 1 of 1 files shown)


kpango (approved):
LGTM

hlts2 (approved):

PR #2172, opened 2023-09-08
Refactor Agent

kpango:
### Description:
SSIA

### Versions:

- Go Version: 1.21.0
- Docker Version: 20.10.8
- Kubernetes Version: v1.28.1
- NGT Version: 2.1.3

### Checklist:


hack/benchmark/internal/operation/search.go
```diff
@@ -167,12 +167,14 @@ func (o *operation) StreamSearchByID(ctx context.Context, b *testing.B, maxIdNum
 		}()
 
 		for i := 0; i < b.N; i++ {
-			err = sc.Send(&payload.Search_IDRequest{
-				Id:     strconv.Itoa(i % maxIdNum),
-				Config: cfg,
-			})
-			if err != nil {
-				b.Fatal(err)
+			if sc != nil {
```
kevindiu, line 170: I think it should be better.

```go
...
if sc != nil {
    for i := 0; i < b.N; i++ {
...
```

internal/core/algorithm/ngt/ngt.go
```diff
@@ -470,14 +476,22 @@ func (n *ngt) LinearSearch(vec []float32, size int) (result []SearchResult, err
 	rsize := int(C.ngt_get_result_size(results, ebuf))
 	if rsize <= 0 {
 		err = n.newGoError(ebuf)
-		if err == nil {
-			err = errors.ErrEmptySearchResult
+		if atomic.LoadUint64(&n.cnt) == 0 {
+			return nil, errors.ErrSearchResultEmptyButNoDataStored
```
kevindiu, line 480: is there any meaning to check the count after the go error is created from ebuf? 
Can we create the go error after the count is checked?

e.g.
```go
...
if atomic.LoadUint64(&n.cnt) == 0 {
	return nil, errors.ErrSearchResultEmptyButNoDataStored
}

err = n.newGoError(ebuf)
if err != nil {
...
```

kpango: Usually rsize = 0 is considered an error, which means that the ebuf needs to be returned to the Pool as soon as possible, and therefore, PutErrorBuffer or newGoError can be called to Put or DestroyCreate the ErrorObject, regardless of the type of error or more. 
In the case of ErrorSearchResultEmptyButNoDataStored, there is no need to return the result of newGoError because the error can occur only in the unindexed case, so no Wrap is also performed.

hack/benchmark/internal/operation/search.go
kpango: Since sc is also used inside the goroutine, I think check sc in all loops just to be safe.
But I'll update the source code for other reasons.

kpango: pushed, please review it again.

vankichi (dismissed):
LGTM

kevindiu (approved):

vankichi (approved):

vankichi (approved):

kevindiu (approved):

PR #2227, opened 2023-11-06
Add implementation for save index job

hlts2:
### Description:

I added a component to execute the `saveIndex` requests to agents.
This component will run as a k8s job in the future.

### NOTE

This component name is `index-save` to match the naming of the other components (`index-correction`, `index-creation`). This may change in the future at the time of refactoring.

### Versions:

- Go Version: 1.21.3
- Docker Version: 20.10.8
- Kubernetes Version: v1.28.2
- NGT Version: 2.1.3

### Checklist:

### DIFF
--- a/cmd/index/job/save/sample.yaml
+++ b/cmd/index/job/save/sample.yaml
@@ -1,0 +1,231 @@
+#
+# Copyright (C) 2019-2023 vdaas.org vald team <vald@vdaas.org>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+version: v0.0.0
+time_zone: JST
+logging:
+  format: raw
+  level: info
+  logger: glg
+server_config:
+  servers:
+    - name: grpc
+      host: 0.0.0.0
+      port: 8081
+      grpc:
+        bidirectional_stream_concurrency: 20
+        connection_timeout: ""
+        header_table_size: 0
+        initial_conn_window_size: 0
+        initial_window_size: 0
+        interceptors: []
+        keepalive:
+          max_conn_age: ""
+          max_conn_age_grace: ""
+          max_conn_idle: ""
+          time: ""
+          timeout: ""
+        max_header_list_size: 0
+        max_receive_message_size: 0
+        max_send_message_size: 0
+        read_buffer_size: 0
+        write_buffer_size: 0
+      mode: GRPC
+      probe_wait_time: 3s
+      restart: true
+  health_check_servers:
+    - name: readiness
+      host: 0.0.0.0
+      port: 3001
+      http:
+        handler_timeout: ""
+        idle_timeout: ""
+        read_header_timeout: ""
+        read_timeout: ""
+        shutdown_duration: 0s
+        write_timeout: ""
+      mode: ""
+      probe_wait_time: 3s
+  metrics_servers:
+  startup_strategy:
+    - grpc
+    - readiness
+  full_shutdown_duration: 600s
+  tls:
+    ca: /path/to/ca
+    cert: /path/to/cert
+    enabled: false
+    key: /path/to/key
+creator:
+  agent_port: 8081
+  agent_name: "vald-agent-ngt"
+  agent_dns: vald-agent-ngt.default.svc.cluster.local
+  agent_namespace: "default"
+  node_name: ""
+  concurrency: 1
+  discoverer:
+    duration: 500ms
+    client:
+      addrs:
+        - vald-discoverer.default.svc.cluster.local:8081
+      health_check_duration: "1s"
+      connection_pool:
+        enable_dns_resolver: true
+        enable_rebalance: true
+        old_conn_close_duration: 3s
+        rebalance_duration: 30m
+        size: 3
+      backoff:
+        backoff_factor: 1.1
+        backoff_time_limit: 5s
+        enable_error_log: true
+        initial_duration: 5ms
+        jitter_limit: 100ms
+        maximum_duration: 5s
+        retry_count: 100
+      call_option:
+        max_recv_msg_size: 0
+        max_retry_rpc_buffer_size: 0
+        max_send_msg_size: 0
+        wait_for_ready: true
+      dial_option:
+        backoff_base_delay: 1s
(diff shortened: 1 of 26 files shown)


pkg/index/job/save/service/indexer_test.go
```diff
@@ -0,0 +1,417 @@
...
+				},
+			}
+		}(),
+		func() test {
+			targetAddrs := []string{
+				"127.0.0.1:8080",
+			}
+			targetAddrList := map[string]bool{
+				targetAddrs[0]: true,
+			}
```
ykadowak, line 158: It seems that targetAddr is registered in targetAddrList here, so why would the test fail?

pkg/index/job/save/usecase/save.go
```diff
@@ -0,0 +1,213 @@
...
+// Start is a method used to initiate an operation in the run, and it returns a channel for receiving errors
+// during the operation and an error representing any initialization errors.
+func (r *run) Start(ctx context.Context) (<-chan error, error) {
+	ech := make(chan error, 4)
+	var sech, oech <-chan error
+	if r.observability != nil {
+		oech = r.observability.Start(ctx)
+	}
+	sech = r.server.ListenAndServe(ctx)
+	ipech, err := r.indexer.PreStart(ctx)
```
ykadowak, line 145: Calling `PreStart` in `Start` might be confusing, so something like `StartMonitoring` should be better?

ykadowak (commented):
Overall looks good. Just some nits.

cmd/index/job/save/sample.yaml
```diff
@@ -0,0 +1,231 @@
...
+  startup_strategy:
+    - grpc
+    - readiness
+  full_shutdown_duration: 600s
+  tls:
+    ca: /path/to/ca
+    cert: /path/to/cert
+    enabled: false
+    key: /path/to/key
+savior:
```
vankichi, line 73:
```suggestion
saver:
```

I think `savior` is NOT much in this context :sweat_smile:

pkg/index/job/save/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+	Server *config.Servers `json:"server_config" yaml:"server_config"`
+
+	// Observability represents observability configurations.
+	Observability *config.Observability `json:"observability" yaml:"observability"`
+
+	// Save represents auto save indexing service configurations.
+	Save *config.IndexSave `json:"savior" yaml:"savior"`
+}
+
+// NewConfig load configurations from file path.
```
vankichi, line 39:
```suggestion
// NewConfig loads configurations from the file path.
```

pkg/index/job/save/service/mock_test.go
```diff
@@ -0,0 +1,57 @@
...
+package service
+
+import (
+	"context"
+
+	"github.com/vdaas/vald/internal/client/v1/client/discoverer"
+	"github.com/vdaas/vald/internal/net/grpc"
+)
+
+type mockDiscovererClient struct {
```
vankichi, line 23: It would be better to move this file to the `internal` directory.
There are mock files in the `internal` package.

pkg/index/job/save/usecase/save.go
```diff
@@ -0,0 +1,213 @@
+// Copyright (C) 2019-2023 vdaas.org vald team <vald@vdaas.org>
```
vankichi, line 1: `saved.go` is better if we should match file naming conventions.

vankichi (changes requested):

pkg/index/job/save/usecase/save.go
ykadowak: This is not a daemon but a job so I think `save.go` is ok for now. We should discuss the naming for all the jobs including this one later when we refactor.

pkg/index/job/save/service/indexer_test.go
hlts2: Sorry for the confusion 🙏 
This test is meant to be a test case for when the address the user expects does not actually exist. 
So GetAddrs returns nil. 👉 https://github.com/vdaas/vald/blob/feature/index-job/save/pkg/index/job/save/service/indexer_test.go#L167]

I'll add a comment as well!

cmd/index/job/save/sample.yaml
hlts2: Thank you for the suggestion 🙏 
I will apply it!

pkg/index/job/save/config/config.go
hlts2: sorry... I will apply it!
Thank you for catching it 🙏

pkg/index/job/save/usecase/save.go
hlts2: @vankichi @ykadowak 
Thank you for your comment!
> saved.go is better if we should match file naming conventions.

Yes, unlike other components, this component does not behave like daemon process. So it has no `d` prefix. 
I think that there is some discussion point about naming, so I will refactor it after all the jobs are implemented. 👍

pkg/index/job/save/service/mock_test.go
hlts2: Thank you for suggestion! 🙏  It looks good to me!
I will create the mock file in `internal/test/mock`

pkg/index/job/save/usecase/save.go
hlts2: Thank you for your suggestion!
How about StartClient? 🤔  since this is actually start of the client rather than monitoring?

pkg/index/job/save/service/indexer_test.go
hlts2: I have fixed! 👉 https://github.com/vdaas/vald/pull/2227/commits/fc44305036bdd13ee29d2c676d73e490ed7cefa3

pkg/index/job/save/service/mock_test.go
hlts2: I have fixed 👉 https://github.com/vdaas/vald/pull/2227/commits/f2dc953c4ccb552287e173e23b15753d531cb3e3

pkg/index/job/save/usecase/save.go
ykadowak: Sounds good.

ykadowak (dismissed):

ykadowak (dismissed):

pkg/index/job/save/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+// NewConfig loads configurations from the file path.
+func NewConfig(path string) (cfg *Data, err error) {
+	cfg = new(Data)
+
+	if err = config.Read(path, &cfg); err != nil {
+		return nil, err
+	}
+
+	if cfg != nil {
+		_ = cfg.GlobalConfig.Bind()
```
vankichi, line 48:
```suggestion
		_ = cfg.Bind()
```

vankichi (changes requested):

vankichi (dismissed):

ykadowak (approved):

vankichi (approved):

PR #2234, opened 2023-11-10
Improve index correction performance

ykadowak:
### Description:

1. Add `GetTimestamp` rpc in agent. This is for reducing network bandwidth by not sending vector data itself, which is not required to correct index. This has a lot of impact when network quality is low.
2. Changed to process agents by the reverse order of index numbers it has. The meaning of this is, by completing the process of a certain agent as soon as possible, it is possible to reduce the number of agents that need to be broadcasted at an early stage, thereby saving network bandwidth and reducing the waiting time for responses. This is effective when there is a bias in the number of indices each agent has.

Overall this PR reduces processing time by 30% compared to the worst case.

### Versions:

- Go Version: 1.21.3
- Docker Version: 20.10.8
- Kubernetes Version: v1.28.2
- NGT Version: 2.1.3

### Checklist:


pkg/agent/core/ngt/handler/grpc/object_test.go
```diff
@@ -1337,6 +1337,118 @@ func Test_server_StreamGetObject(t *testing.T) {
 	}
 }
 
+func Test_server_GetTimestamp(t *testing.T) {
```
hlts2, line 1340: Is there a reason you are implementing this test without using a vald test template? 🤔

pkg/agent/core/ngt/handler/grpc/object_test.go
```diff
@@ -1337,6 +1337,118 @@ func Test_server_StreamGetObject(t *testing.T) {
...
+		BrokenIndexHistoryLimit: 1,
+	}
+
+	setup := func(t *testing.T) (context.Context, Server) {
+		t.Helper()
+		ngt, err := service.New(&defaultConfig)
+		require.NoError(t, err)
+
+		ctx := context.Background()
+		eg, ectx := errgroup.New(ctx)
```
hlts2, line 1364: I think it is better to execute eg.Wait in each test case with just in case 🤔

pkg/agent/core/ngt/handler/grpc/object_test.go
```diff
@@ -1337,6 +1337,118 @@ func Test_server_StreamGetObject(t *testing.T) {
...
+
+	type test struct {
+		name     string
+		testfunc func(t *testing.T)
+	}
+
+	tests := []test{
+		{
+			name: "succeeds to get object meta",
+			testfunc: func(t *testing.T) {
```
hlts2, line 1384: I think It is better to separate the setup logic(`Insert, CreateIndex`..etc) for testing from the comparison by actual testing to improve readability and maintainability. 🤔

pkg/agent/core/ngt/handler/grpc/object_test.go
ykadowak: Because in the handler test, not only the arguments of the function call, but also the flow of the preliminary setup is part of the test. We have already agreed on this, and some tests in this file are implemented in the same way.

pkg/agent/core/ngt/handler/grpc/object_test.go
ykadowak: [2b0715b](https://github.com/vdaas/vald/pull/2234/commits/2b0715bf2685ab24d470fb7714dc20b221339d1c)

pkg/agent/core/ngt/handler/grpc/object_test.go
ykadowak: With the same reason as above, this preliminary setup is also part of the test so I don't think it should be separated. But I added a comment to make it clearer.
[6e3c150](https://github.com/vdaas/vald/pull/2234/commits/6e3c150fae457b1ec9578a47961711d8c500406b)

datelier (approved):

hlts2 (approved):
LGTM

PR #2263, opened 2023-12-07
Add svc discoverer for readreplica svc

ykadowak:
### Description:

This PR

~~1.  Adds `internal/k8s/readreplica/svc` to reconcile readreplica svc resource~~
1. Adds `internal/k8s/svc` to reconcile svc resources
2. Updates discoverer to collect readreplica svc addrs and replica id by using the reconciler

### Versions:

- Go Version: 1.21.5
- Docker Version: 20.10.8
- Kubernetes Version: v1.28.4
- NGT Version: 2.1.5

### Checklist:


vankichi:
IMO, it would be better to change the path from `internal/k8s/readreplica/svc` to `internal/k8s/svc`, because it can be used for any methods not only for readReplica.

ykadowak:
> IMO, it would be better to change the path from `internal/k8s/readreplica/svc` to `internal/k8s/svc`, because it can be used for any method not only for readReplica.

I was thinking of that but this controller has to know `readreplica` specific things like `replica id` so it cannot be used generally.

internal/k8s/readreplica/svc/svc.go
```diff
@@ -0,0 +1,160 @@
...
+				svc.GetName(),
+				svc.Status,
+				svc.GetDeletionTimestamp())
+			continue
+		}
+		labels := svc.GetLabels()
+		v, ok := labels[r.idKey]
+		if !ok {
+			log.Errorf("this svc(%s) does not have readreplica id label(%s)", svc.GetName(), r.idKey)
+			return reconcile.Result{}, fmt.Errorf("no valid label is put in the svc")
```
vankichi, line 118:
```suggestion
			return reconcile.Result{}, errors.New("no valid label is put in the svc")
```
If you do not need to format, `errors.New()` is better for performance.

vankichi (changes requested):

vankichi:
> > IMO, it would be better to change the path from `internal/k8s/readreplica/svc` to `internal/k8s/svc`, because it can be used for any method not only for readReplica.
> 
> I was thinking of that but this controller has to know `readreplica` specific things like `replica id` so it cannot be used generally.

Hmm...
In the current implementation, a specific `replica id` is used for checking whether labels have replica key id, so how about changing the variable name in the struct from `keyId string` to `labels []string` and changing checking logic?
It is a just idea, and I'm wondering which is best, so I would like to hear your and other members' opinions.

ykadowak:
@vankichi 
Moved `internal/k8s/readreplica/svc` to `internal/k8s/svc` as we discussed by passing `corev1.Service` to `onReconcile`.
[2725af5](https://github.com/vdaas/vald/pull/2263/commits/2725af5d3133bc8f0b0373864dc79d59bce64065)

vankichi (approved):
Thanks! LGTM

vankichi (commented):
Overall LGTM.
I would appreciate it if you added a comment for each Global value/function that you implemented.
e.g.
    - `internal/k8s/service/...` files 
    - `pkg` files

vankichi (approved):
LGTM

kpango (approved):
LGTM

PR #2298, opened 2024-01-19
Add E2E tests for read replica feature

ykadowak:
### Description:

This PR adds E2E for read replica as `TestE2EReadReplica`.
- Add minikube as e2e env because read replica needs `ROX` policy that k3d does not support.
- Update some e2e k8s client helper
- Implement the e2e test logic
  - deploy vald and vald-readreplica -> insert -> save index -> rotate read replica -> search

### Versions:

- Go Version: 1.21.5
- Docker Version: 20.10.8
- Kubernetes Version: v1.28.4
- NGT Version: 2.1.6

### Checklist:

### DIFF
--- a/tests/e2e/crud/crud_test.go
+++ b/tests/e2e/crud/crud_test.go
@@ -827,7 +827,7 @@ func TestE2EIndexJobCorrection(t *testing.T) {
 
 func TestE2EReadReplica(t *testing.T) {
 	t.Cleanup(teardown)
-	
+
 	if kubeClient == nil {
 		var err error
 		kubeClient, err = client.New(kubeConfig)


tests/e2e/crud/crud_test.go
```diff
@@ -104,14 +106,12 @@ func init() {
 	flag.Parse()
 
 	var err error
+	kubeClient, err = client.New(*kubeConfig)
```
hlts2, line 109: In this case, is it necessary that the environment in which the test is run always have information about the connection to the Kubernetes cluster? 🤔

vankichi:
I will LGTM if the above conversation has been resolved

tests/e2e/crud/crud_test.go
ykadowak: Yes.

`TestE2EReadReplica` needs to be dependent on k8s client anyway so I guess I should revert this and check this in `TestE2EReadReplica`? That way other e2e tests can run without k8s clinet.

ykadowak: I fixed that way [ed4e8b3](https://github.com/vdaas/vald/pull/2298/commits/ed4e8b3ec09da7f635624ec3110626b39a676e49) since I think it's better anyway.

Let me know if this is not what you intended.

hlts2: > so I guess I should revert this and check this in TestE2EReadReplica?

Thank you for reply :pray:
Yes, I think that would be better!!

hlts2: I reviewed your changes! It looks good to me! 🙏 
Thank you 🙏

vankichi (approved):

PR #2352, opened 2024-02-15
create continous benchmark doc

vankichi:
### Description

This PR has 2 changes:

- Add deploy continuous benchmark operator
- Remove `dimension` field from `ValdBenchmarkJobRelease`

#### AsIs
It is used only search operation for checking whether search query's dimension is correct in current.
It is involved to dataset in the above sense, but this field is not in dataset object.

#### ToBe
I think we have two options:

1. move dimension field into the dataset object
1. remove dimension field

1 is not bad. But, I don't think there is any need to set it in the dimension field because you can check first whether dimension of the dataset matches dimension set in Vald cluster before apply manifest.
About 2, it would reduce the number of parameters to be set, and in any case an error would occur if the number of dimensions was incorrect.

### Versions

- Go Version: 1.22.0
- Docker Version: 20.10.8
- Kubernetes Version: v1.29.1
- NGT Version: 2.1.6

### Checklist


docs/performance/continuos-benchmatk.md
```diff
@@ -0,0 +1,280 @@
+# Continuous Benchmark Tool
+
+## What is the Continuous Benchmark Tool?
+
+Continuous Benchmark Tool allows you to get benchmark of Vald cluster in 24/365.
```
kpango, line 5:
```suggestion
Continuous Benchmark Tool allows you to get benchmark of Vald cluster in 24/7.
```

kpango (changes requested):
Can you please explain why the dimension has been removed in this PR?

vankichi:
> Can you please explain why the dimension has been removed in this PR?

As I asked main contributors, `dimension` is not necessary.

## AsIs

It is used only search operation for checking whether search query's dimension is correct in current.
It is involved to dataset in the above sense, but this field is not in `dataset` object.

## ToBe

I think we have two options:
1. move `dimension` field into the `dataset` object
1. remove `dimension` field

1 is not bad. But, I don't think there is any need to set it in the dimension field because you can check first whether dimension of the dataset matches dimension set in Vald cluster before apply manifest.
About 2, it would reduce the number of parameters to be set, and in any case an error would occur if the number of dimensions was incorrect.

kpango (approved):
LGTM

docs/performance/continuos-benchmatk.md
```diff
@@ -0,0 +1,280 @@
+# Continuous Benchmark Tool
+
+## What is the Continuous Benchmark Tool?
+
+Continuous Benchmark Tool allows you to get benchmark of Vald cluster in 24/7.
+
+Assumed use case is:
+
+- Verification with workload close to production environment
```
hlts2, line 9:
```suggestion
- Verification with workload close to the production environment
```

hlts2 (approved):

kpango (approved):

PR #2428, opened 2024-02-28
Added snapshot timestamp annotations to read replica agent

ykadowak:
### Description

- Added a timestamp of the snapshot taken from the primary agent to the annotations of the read replica agent.
- Added logic to the index operator to verify the need for rotation based on the timestamp mentioned above.

### Versions

- Go Version: 1.22.0
- Docker Version: 20.10.8
- Kubernetes Version: v1.29.2
- NGT Version: 2.1.6

### Checklist


kmrmt (dismissed):

kpango (changes requested):

internal/k8s/pod/pod.go
```diff
@@ -110,7 +111,8 @@ func (r *reconciler) Reconcile(ctx context.Context, _ reconcile.Request) (res re
 		pods       = make(map[string][]Pod, len(ps.Items))
 	)
 
-	for _, pod := range ps.Items {
+	for i := range ps.Items {
+		pod := ps.Items[i]
```
kpango, line 115: Can you please describe this change?

pkg/index/operator/service/operator.go
```diff
@@ -127,15 +141,79 @@ func (o *operator) podOnReconcile(ctx context.Context, podList map[string][]pod.
...
 	}
 }
 
 // TODO: implement job reconcile logic to detect save job completion and to start rotation.
-func (o *operator) jobOnReconcile(ctx context.Context, jobList map[string][]job.Job) {
+func (*operator) jobOnReconcile(_ context.Context, jobList map[string][]job.Job) {
 	for k, v := range jobList {
-		for _, job := range v {
+		for i := range v {
+			job := v[i]
```
kpango, line 159: This change too

internal/k8s/pod/pod.go
ykadowak: To avoid copy to fix the deepsource error.

pkg/index/operator/service/operator.go
ykadowak: same as above

internal/k8s/pod/pod.go
kpango: I think Range Over Copy value is less of a problem since index access to slice causes BoundaryOverflow problems.
How do you think?

ykadowak: I don't mind reverting this. In that case, I think we should turn off the deepsource error by default.

kpango: yes exactly can you add skipcq annotation for supress deepsource?

ykadowak: I was thinking of turning it off globally but deepsource does not offer the feature...

[2cc4c84](https://github.com/vdaas/vald/pull/2428/commits/2cc4c8483dd99c617758454d7d2f6b3848e60543)

kpango: thank you for revert and add annotations.

kpango (approved):
LGTM

kmrmt (approved):

PR #2438, opened 2024-03-05
Fix index job logic to pass DNS A record

hlts2:
### Description

In the previous implementation, it was possible to set a DNS A record to a job.
However, since only IP was enabled internally, even if a DNS A record was set, it could not be properly connected.
Therefore, I will modify the logic so that when a user specifies a target address (`A record:port` or `ip:port`), it will be preferentially selected.

### Versions

- Go Version: 1.22.0
- Docker Version: 20.10.8
- Kubernetes Version: v1.29.2
- NGT Version: 2.1.6

### Checklist


hlts2:
> **NOTE**
The operator e2e is falling, but it is not falling due to this changes. I will create a new PR for this issue.
https://github.com/vdaas/vald/actions/runs/8227057117/job/22495966605

ykadowak (approved):

kmrmt (approved):

PR #2444, opened 2024-03-08
Implement index operator logic for read replica rotation

ykadowak:
### Description

This PR implements the reconcile logic for read replica rotator in index operator.

- It retrieves the last save and rotation timestamps from the annotations of the agent and the read replica agent. If the last save happened after the last rotation, it generates a rotator job.
- If there are more rotation jobs running than the specified `rotationJobConcurrency`, it does not create a new job but requeues instead.

Will Implement tests after implementation fixed.

### Versions

- Go Version: 1.22.0
- Docker Version: 20.10.8
- Kubernetes Version: v1.29.2
- NGT Version: 2.1.6

### Checklist


charts/vald/templates/index/job/readreplica/rotate/configmap.yaml
```diff
@@ -48,6 +48,6 @@ data:
     rotator:
       agent_namespace: {{ $rotator.agent_namespace | quote }}
       read_replica_label_key: {{ $agent.readreplica.label_key | quote }}
-      read_replica_id: "_MY_TARGET_REPLICA_ID_"
+      read_replica_id: "_{{ $rotator.target_read_replica_id_envname}}_"
```
kpango, line 51:
```suggestion
      read_replica_id: "{{ $rotator.target_read_replica_id }}"
```

charts/vald/values.yaml
```diff
@@ -3475,9 +3475,9 @@ manager:
         # @schema {"name": "manager.index.readreplica.rotator.agent_namespace", "type": "string"}
         # manager.index.readreplica.rotator.agent_namespace -- namespace of agent pods to manage
         agent_namespace: _MY_POD_NAMESPACE_
-        # @schema {"name": "manager.index.readreplica.rotator.read_replica_id", "type": "string"}
-        # manager.index.readreplica.rotator.read_replica_id -- read replica id to perform rotation
-        read_replica_id: _MY_TARGET_REPLICA_ID_
+        # @schema {"name": "manager.index.readreplica.rotator.target_read_replica_id_envname", "type": "string"}
+        # manager.index.readreplica.rotator.target_read_replica_id_envname -- read replica id to perform rotation
+        target_read_replica_id_envname: MY_TARGET_REPLICA_ID
```
kpango, line 3480:
```suggestion
        # @schema {"name": "manager.index.readreplica.rotator.target_read_replica_id", "type": "string"}
        # manager.index.readreplica.rotator.target_read_replica_id -- read replica id to perform rotation
        target_read_replica_id: _MY_TARGET_REPLICA_ID_
```

internal/config/index_operator.go
```diff
@@ -15,14 +15,23 @@ package config
...
 	// AgentNamespace represent agent namespace location
 	AgentNamespace string `json:"agent_namespace" yaml:"agent_namespace"`
 
-	// Concurrency represents indexing concurrency.
-	Concurrency int `json:"concurrency" yaml:"concurrency"`
+	// RotatorName represent rotator name for service discovery
+	RotatorName string `json:"rotator_name" yaml:"rotator_name"`
+
+	// TargetReadReplicaIDEnvname represents the environment variable name for target read replica id.
+	TargetReadReplicaIDEnvname string `json:"target_read_replica_id_envname" yaml:"target_read_replica_id_envname"`
```
kpango, line 31:
```suggestion
	// TargetReadReplicaID represents the target read replica id.
	TargetReadReplicaID string `json:"target_read_replica_id" yaml:"target_read_replica_id"`
```

internal/config/index_operator.go
```diff
@@ -32,6 +41,7 @@ type IndexOperator struct {
 }
 
 func (ic *IndexOperator) Bind() *IndexOperator {
+	ic.Namespace = GetActualValue(ic.Namespace)
```
kpango, line 44:
```suggestion
	ic.Namespace = GetActualValue(ic.Namespace)
	ic.TargetReadReplicaID = GetActualValue(ic.TargetReadReplicaID)
```

kpango (changes requested):
Since the Config struct is designed to receive values directly, explicitly specifying that it is an Envvvar is not necessary, and the only case where the value passed from yaml is a special case (_), so don't care about it.

charts/vald-helm-operator/crds/valdrelease.yaml
```diff
@@ -11240,6 +11238,8 @@ spec:
                                       type: boolean
                                     name:
                                       type: string
+                                target_read_replica_id_envname:
```
hlts2, line 11241:
```suggestion
                                target_read_replica_id:
```

charts/vald/templates/index/operator/configmap.yaml
```diff
@@ -42,9 +43,12 @@ data:
       {{- $observability := dict "Values" $operator.observability "default" .Values.defaults.observability }}
       {{- include "vald.observability" $observability | nindent 6 }}
     operator:
+      namespace: _MY_POD_NAMESPACE_
       agent_name: {{ $agent.name }}
       agent_namespace: {{ $agent.namespace }}
-      concurrency: 1
+      rotator_name: {{ $rotator.name }}
+      target_read_replica_id_envname: {{ $rotator.target_read_replica_id_envname }}
```
hlts2, line 50:
```suggestion
      target_read_replica_id: {{ $rotator.target_read_replica_id }}
```

cmd/index/operator/sample.yaml
```diff
@@ -69,8 +69,11 @@ server_config:
     enabled: false
     key: /path/to/key
 operator:
+  namespace: "default"
   agent_name: "vald-agent"
   agent_namespace: "default"
+  rotator_name: "vald-readreplica-rotate"
+  target_read_replica_id_envname: MY_TARGET_REPLICA_ID
```
hlts2, line 76:
```suggestion
  target_read_replica_id: _MY_TARGET_REPLICA_ID_
```

ykadowak:
> Since the Config struct is designed to receive values directly, explicitly specifying that it is an Envvvar is not necessary, and the only case where the value passed from yaml is a special case (_), so don't care about it.

This is to specify which read replica to rotate on creating rotator job like below. I made the envname variable to avoid hard coding the envname.

https://github.com/vdaas/vald/blob/d4a19aa2c077514b82bce571484e2fdc2ed26936/pkg/index/operator/service/operator.go#L260-L263

kpango:
> > Since the Config struct is designed to receive values directly, explicitly specifying that it is an Envvvar is not necessary, and the only case where the value passed from yaml is a special case (_), so don't care about it.
> 
> @kpango This is to specify which read replica to rotate on creating rotator job like below. I made the envname variable to avoid hard coding the envname.
> 
> https://github.com/vdaas/vald/blob/d4a19aa2c077514b82bce571484e2fdc2ed26936/pkg/index/operator/service/operator.go#L260-L263

Yes, I know that, so I think you can do what you are supposed to do if you modify as I suggested.

internal/k8s/podv2/option.go
```diff
@@ -0,0 +1,96 @@
...
+			r.addListOpts(client.MatchingFields(fs))
+		}
+		return nil
+	}
+}
+
+func WithForOpts(fopts ...builder.ForOption) Option {
+	return func(r *reconciler) error {
+		r.forOpts = fopts
+		return nil
```
hlts2, line 94: If called more than once, the options added before will disappear, so I think it is better to use append just in case 👍 
```suggestion
	return func(r *reconciler) error {
		if len(fopts) == 0 {
			return nil
		}
		r.forOpts = append(r.forOpts, fopts...)
		return nil
```

kpango (approved):

internal/k8s/client/client.go
```diff
@@ -49,13 +54,21 @@ type (
...
 	MatchingLabels     = cli.MatchingLabels
 	InNamespace        = cli.InNamespace
 	VolumeSnapshot     = snapshotv1.VolumeSnapshot
+	Pod                = corev1.Pod
 	Deployment         = appsv1.Deployment
 	DeploymentList     = appsv1.DeploymentList
 	ObjectMeta         = metav1.ObjectMeta
+	EnvVar             = corev1.EnvVar
+	Job                = batchv1.Job
+	JobList            = batchv1.JobList
```
vankichi, line 69: Please explain why define `Job` and `JobList` here.
`Job` is already at defined `internal/k8s/job/job`

vankichi (approved):

PR #2454, opened 2024-03-15
add inner product distance type for ngt

kpango:
### Description

I've added NGT's InnerProduct distance type to Vald Agent.

### Versions

- Go Version: 1.22.1
- Docker Version: 20.10.8
- Kubernetes Version: v1.29.2
- NGT Version: 2.1.6

### Checklist


ykadowak (approved):
It would be better to have a test to check it a search succeeds with all the distance type settings

kpango:
> It would be better to have a test to check if a search succeeds with all the distance type settings

Exactly, but for now we don't have such datasets for all distance type.
Let's add our future task about it.

kmrmt (approved):
If I remember correct, inner product supports only one shot create_index. Do we have manual operation temporary and will controlling the one shot be implemented another PR later?

PR #2797, opened 2025-01-08
Fix validation logic for manager indexer

aknishid:
### Description
It appears that the current validation does not account for cases where saveIndexDurationLimit might be set to a significantly large negative value (e.g., -1h). This could inadvertently result in valid configurations being flagged as errors.

Even when indexDuration, indexDurationLimit, and saveIndexDurationLimit are all set to negative values, the system assigns MaxInt64 at startup, ensuring that an index creation will eventually occur. Because of this, I wonder if this validation is strictly necessary.

If the goal is to detect only configurations where index duration is effectively disabled, we could modify the validation to:
```
if i.indexDuration <= 0 && i.indexDurationLimit <= 0 && i.saveIndexDurationLimit <= 0 {
    return nil, errors.ErrInvalidConfig
}

```
This way, the validation triggers an error only when all three parameters are negative.

As I do not have full context on the original intent behind this validation, I would appreciate any insights or corrections if I am misunderstanding its purpose.

### Versions

- Vald Version: v1.7.15
- Go Version: v1.23.4
- Rust Version: v1.83.0
- Docker Version: v27.4.0
- Kubernetes Version: v1.32.0
- Helm Version: v3.16.3
- NGT Version: v2.3.5
- Faiss Version: v1.9.0

### Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved validation logic for configuration duration parameters to ensure each duration is non-negative.


kpango (dismissed):
LGTM

kpango (approved):

vankichi:
Thanks @aknishid 
I will approve after all e2e-test on GitHub Actions end with success.

PR #2803, opened 2025-01-16
refactor: Use generics in function of bidirectional stream

highpon:
### Description
Use generics of function of bidirectional stream.
This PR intends to refactor current code.

### Versions

- Vald Version: v1.7.15
- Go Version: v1.23.4
- Rust Version: v1.83.0
- Docker Version: v27.4.0
- Kubernetes Version: v1.32.0
- Helm Version: v3.16.3
- NGT Version: v2.3.5
- Faiss Version: v1.9.0

### Checklist

## Summary by CodeRabbit

- **Refactor**
	- Updated function signatures in multiple service files to improve type safety.
	- Modified return types to use pointer-based generics in gRPC stream and load testing utilities.
	- Enhanced type handling for data providers and request generation.

- **Technical Improvements**
	- Refined type parameter declarations in gRPC streaming functions.
	- Updated callback and data provider function signatures to use more explicit pointer types.

### DIFF
--- a/internal/net/grpc/stream.go
+++ b/internal/net/grpc/stream.go
@@ -218,7 +218,7 @@ func BidirectionalStreamClient[S, R any](
 					return eg.Wait()
 				}
 
-				err = stream.SendMsg(data)
+				err = stream.SendMsg(*data)
 				if err != nil {
 					return err
 				}
--- a/internal/net/grpc/stream.go
+++ b/internal/net/grpc/stream.go
@@ -45,7 +45,7 @@ type (
 // It receives messages from the stream, calls the function with the received message, and sends the returned message to the stream.
 // It limits the number of concurrent calls to the function with the concurrency integer.
 // It records errors and returns them as a single error.
-func BidirectionalStream[Q any, R any](
+func BidirectionalStream[Q, R any](
 	ctx context.Context,
 	stream ServerStream,
 	concurrency int,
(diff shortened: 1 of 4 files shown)


kpango:
@highpon Thank you for your contribution around important libs.

How about an interface that cancels all stream processing in case of false by using a boolean as the return value of the callback function argument of the BidirectionalStreamClient?

It would be very useful to have an interface that can stop all streams when a fatal error is detected in the callback function.

internal/net/grpc/stream.go
```diff
@@ -167,8 +167,8 @@ func BidirectionalStream[Q any, R any](
 }
 
 // BidirectionalStreamClient is gRPC client stream.
-func BidirectionalStreamClient(
-	stream ClientStream, dataProvider, newData func() any, f func(any, error),
+func BidirectionalStreamClient[S, R any](
+	stream ClientStream, sendDataProvider func() *S, callBack func(*R, error),
```
kpango, line 171: @highpon Thank you for your contribution around important libs.

How about an interface that cancels all stream processing in case of false by using a boolean as the return value of the callback function argument of the BidirectionalStreamClient?

It would be very useful to have an interface that can stop all streams when a fatal error is detected in the callback function.

```suggestion
func BidirectionalStreamClient[S, R any](
	stream ClientStream, sendDataProvider func() *S, callBack func(*R, error) bool,
```

internal/net/grpc/stream.go
```diff
@@ -183,13 +183,13 @@ func BidirectionalStreamClient(
...
 			default:
-				res := newData()
+				res := new(R)
 				err = stream.RecvMsg(res)
 				if err == io.EOF || errors.Is(err, io.EOF) {
 					cancel()
 					return nil
 				}
-				f(res, err)
+				callBack(res, err)
```
kpango, line 192: Kind of like this.
```suggestion
				if !callBack(res, err) {
					cancel()
					return nil
				}
```

highpon: Thanks for you review!
fix: https://github.com/vdaas/vald/pull/2803/commits/39c214947bb034337a86aacb33e84e9256311d6d

kmrmt (approved):

kpango (approved):

PR #3520, opened 2026-04-07
fix: resolve docker/CI build failures from cmake flag mixing and toolchain export pollution

kpango:
Docker builds fail because `LDFLAGS` (containing `-static`, `-Wl,--whole-archive`, `-fuse-ld=lld`, etc.) was passed into `CMAKE_C_FLAGS`/`CMAKE_CXX_FLAGS`, causing cmake compiler checks to fail. Additionally, `export CC CXX AR NM RANLIB LDFLAGS CFLAGS CXXFLAGS` polluted subprocess environments, breaking `cargo install` (ring crate) in format CI where clang isn't installed.

### cmake flag separation
- `CMAKE_C_FLAGS` now gets only compile flags (`$(CFLAGS) -flto=thin`), linker flags go to `CMAKE_EXE_LINKER_FLAGS`
- Removed hardcoded `-DOpenMP_*_LIB_NAMES="gomp"` — let cmake's FindOpenMP handle discovery
- Applied same fix to faiss cmake configuration

### toolchain variable handling
- `CC ?= clang` instead of `CC = clang` (and CXX/AR/NM/RANLIB) — allows CI environments to override
- `export SUDO` only — removed global export of CC/CXX/AR/NM/RANLIB/LDFLAGS/CFLAGS/CXXFLAGS; these are already passed explicitly per-command in `go-build`, cmake invocations, etc.

### portability
- `-fopenmp` instead of `-fopenmp=libgomp` (toolchain-agnostic, works with both GCC and clang)
- `NGT_EXTRA_CMAKE_FLAGS ?=` empty default instead of `-DNGT_AVX2=ON`
- Removed `target-cpu=native` from `rust/.cargo/config.toml`


kpango:
@copilot+claude-opus-4.6 Can you please fix docker build failure?

PR #3531, opened 2026-06-11
[Bugfix] Fix SIGABRT in NGT Agent via runtime.Pinner

kpango:
## Summary

Fix a SIGABRT / memory-corruption crash in the NGT agent that occurred when Go's garbage collector moved a slice's backing array while the NGT C library was still reading from it during search and insert operations. The fix uses `runtime.Pinner` (Go 1.21+) to pin Go memory for the duration of each CGo call. Additionally, a shared Rust build-utils crate is extracted to deduplicate OpenMP detection logic between the `ngt` and `qbg` build scripts, and a context leak in the Docker generator is resolved.

## Background / Motivation

The original code called `vec = nil` / `cvec = nil` immediately after handing Go slice pointers to the NGT C API. This was intended as a cleanup, but it had the opposite effect: it removed the only Go-visible references to the backing arrays, allowing the GC to collect or move the memory while C code was still reading from it. The result is a SIGABRT or silent memory corruption under concurrent load.

`runtime.Pinner` is the correct mechanism: it prevents the GC from moving or releasing pinned objects until `Unpin()` is called. Since Go 1.21, it is the idiomatic way to keep CGo-passed memory stable.

A secondary issue on Ubuntu 25+ is that NGT is compiled with clang's `-fopenmp=libomp`, which emits `__kmpc_*` symbols instead of the `GOMP_*` symbols from GCC's libgomp. Hard-coding `cargo:rustc-link-lib=dylib=gomp` in both `ngt` and `qbg` build scripts caused linker failures on those systems. The fix inspects the static archive with `nm` at build time to choose the correct runtime.

## Changes

- **`internal/core/algorithm/ngt/ngt.go`**: Extract `pinVec()` helper that pins `vec[0]` with `runtime.Pinner`; add `defer pinVec(vec).Unpin()` in `Search`, `LinearSearch`, and `Insert`; remove the unsafe `vec = nil` / `cvec = nil` nulling; update CGo `LDFLAGS` to include `-L/usr/local/lib`; modernise integer-range `for` loops to `for i := range n` form
- **`rust/libs/build-utils/`**: New `vald-build-utils` crate with `link_openmp()` that detects LLVM vs GCC OpenMP at link time by inspecting `libngt.a` with `nm`, probing `/usr/lib/llvm-{14..25}/lib` for `libomp.so`
- **`rust/libs/algorithms/ngt/build.rs`, `rust/libs/algorithms/qbg/build.rs`**: Replace hard-coded `cargo:rustc-link-lib=dylib=gomp` with `vald_build_utils::link_openmp("/usr/local/lib/libngt.a")`
- **`rust/libs/algorithms/qbg/src/lib.rs`**: Replace sequential ID equality assertions with uniqueness checks via `assert_unique_id()` helper; ID ordering after `build_index` is not guaranteed
- **`rust/bin/agent/src/handler/common.rs`**: Extract `get_hostname()` function replacing the removed `cargo::util::hostname`; move all `use` imports above function definitions per Rust convention; add unit tests for env-var, empty-env, and no-env paths
- **`hack/docker/gen/main.go`**: Call `cancel()` before `log.Fatal()` to prevent context resource leak (DeepSource CRT-D0011); add `llvm` to `clangLTOBuildDeps`
- **`internal/core/algorithm/ngt/ngt_test.go`**: Simplify redundant `nil || len == 0` checks to `len == 0`
- **`.github/workflows/unit-test-rust.yaml`**: Install `libngt.a` manually before Rust QBG tests; add NGT cmake build with LTO disabled to avoid `ar`/`rust-lld` mismatch on CI

## Technical Details

- `pinVec` pins only `&vec[0]`; the Go runtime guarantees the remainder of the slice follows the first element, so a single pin suffices.
- `runtime.Pinner.Unpin()` is called via `defer` so it always fires even if the CGo call panics through a Go wrapper.
- `InsertCommit` and `BulkInsert` delegate to `Insert`, which already pins — no additional pinning needed at those call sites.
- The `link_openmp` heuristic falls back to `libgomp` if `nm` is unavailable, preserving compatibility with build environments that lack binutils.
- The `vald-build-utils` crate has no runtime dependencies (build-only), keeping the final binary unaffected.
- `#[allow(deprecated)]` is avoided: `std::env::set_var` in tests is called inside `unsafe` with a safety comment noting single-threaded test context.

## Testing / Verification

- CI Rust unit tests: `unit-test-rust.yaml` updated to build and install NGT before running QBG tests
- Go unit tests: `go test ./internal/core/algorithm/ngt/...` covers Search, LinearSearch, Insert paths with the new Pinner
- QBG insert ID assertion: relaxed to uniqueness check to accommodate SIMD-disabled `build_index` ID reordering; prevents false-negative failures on CI runners without AVX
- `get_hostname()`: three unit tests covering env-set, empty-env fallback, and no-env fallback paths

## Vald Law Compliance

- Law 1 (No pb.go edits): ✅ Not applicable — no protobuf generated files modified
- Law 2 (No panic/log.Fatal): ✅ Compliant — `log.Fatal` in `hack/docker/gen/main.go` is pre-existing generator code (not production service code); `cancel()` is now called before it
- Law 3 (No _ = err): ✅ Compliant
- Law 4 (No stdlib log): ✅ Compliant — no stdlib `log` added to service code
- Law 5 (No secrets): ✅ Compliant

## Related PRs

Part of the NGT SIGABRT fix series: **#3531** → #3532 → #3533 → #3534 → #3535 → #3536 → #3537

This PR is the base: it contains the `runtime.Pinner` core fix, shared Rust build infrastructure, and CI stabilisation. Subsequent PRs in the series build on top of this foundation.


rust/libs/build-utils/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+/// LLVM version range to probe when searching for libomp.so.
+/// Updated as new LLVM releases ship distro packages.
+const LLVM_VERSION_MIN: u32 = 14;
+const LLVM_VERSION_MAX: u32 = 25;
```
Matts966, line 20: Is this fixed upper bound requirement necessary for the build?

Matts966 (approved):

datelier (approved):

PR #3560, opened 2026-07-11
fix(vald-operator): bring Helm chart CI/Makefile parity with vald-helm-operator and vald-benchmark-operator

kpango:
## Summary
- `vald-operator` (charts/operator/vald) was never wired into the Helm publish/lint/release automation, so it was never packaged and never appeared in the published Helm repo index (verified against gh-pages `charts/index.yaml`).
- Added `helm/package/all`, `helm/docs/all` (helm.mk) and `k8s/manifest/all` (k8s.mk) aggregate Makefile targets mirroring the existing `helm/schema/all` idiom, so future charts only need to be listed in one place.
- `helm.yaml`: package all 5 charts instead of 4 — fixes the ArtifactHub publishing gap for vald-operator.
- `helm-lint.yaml`: collapsed 3 near-duplicate per-chart lint jobs into a matrix job and added previously-missing lint coverage for vald-benchmark-operator.
- `_release-pr.yaml` / `update-deps.yaml`: bump vald-operator's chart version and regenerate its k8s manifests/docs on release and daily dependency updates, same as the other charts; also stage `pkg/` since `k8s/manifest/all` now regenerates vald-operator's Go test golden file.
- `k8s.mk`: fixed `mv` path bugs in `k8s/manifest/operator/helm/update`, `k8s/manifest/operator/benchmark/update`, `k8s/operator/helm/deploy`, `k8s/operator/helm/delete`, `k8s/operator/benchmark/deploy`, `k8s/operator/benchmark/delete` — `helm template --output-dir` names the output directory after `Chart.yaml`'s `name:` field (`vald-helm-operator`, `vald-benchmark-operator`), not the chart's source path, so these targets always failed before this fix.

## Verification
- `make helm/package/all` runs end-to-end, producing all 5 chart packages including `vald-operator-*.tgz`.
- `helm lint` passes on all 4 charts (vald, vald-helm-operator, vald-operator, vald-benchmark-operator).
- Regenerated `k8s/operator/helm` and `k8s/operator/benchmark` manifests with the project-pinned helm v4.1.3, ran the standard `make license` + `yamlfmt` cleanup pass (same as CI does downstream), and confirmed the output byte-matches the committed manifests.
- `actionlint` clean on all 4 modified workflow files; YAML syntax validated.
- Independently reviewed by `code-reviewer` and `vald-reviewer` subagents; both flagged the same `pkg/` git-add gap (fixed) and `code-reviewer` additionally caught the same mv-path bug in the `k8s/operator/{helm,benchmark}/deploy|delete` dev-convenience targets (fixed).

## Known/out-of-scope items (confirmed with requester)
- `charts/operator/vald/crds/valdoperatorrelease.yaml` is a static controller-gen artifact with no regeneration mechanism left after the kubebuilder removal — flagged but not fixed here (owned by the operator's Go API type design, out of scope for this CI/Helm parity fix).
- ArtifactHub's `artifacthub-repo.yml` lives on the `gh-pages` branch (repo ownership/verification metadata) — out of scope for this branch.

## Manual follow-up required before/after merge
- `helm-lint.yaml`'s job IDs changed from `lint-vald-chart` / `lint-vald-helm-operator-chart` / `lint-vald-operator-chart` to a single matrix job `lint-chart`. If GitHub branch protection on this or downstream branches lists the old job names as required status checks, please update them post-merge.

Generated with [Claude Code](https://claude.ai/claude-code)


kpango:
Follow-up commit: introduced `hack/helm/charts.yaml` as a single source of truth for the chart list consumed by `helm-lint.yaml`'s lint matrix and `_release-pr.yaml`'s version-bump step, replacing the hardcoded chart array / 8 sed lines respectively. Adding a future chart now only requires one entry in that file instead of editing two CI workflows.

Verified: `yq` queries reproduce prior hardcoded behavior exactly (incl. the `vald-readreplica`/`vald` values.yaml symlink case), `resolve-charts` runs inside the same pinned `vald-dev-container` as every other job (no second yq-provisioning path), `actionlint` clean, and matrix/loop values are passed via `env:` rather than raw `${{ }}` interpolation in `run:` blocks per code review.

Independently reviewed by `code-reviewer` and `vald-reviewer` subagents; both raised no blocking issues. One raised finding (resolve-charts should run in the pinned container, not via a separate `setup-yq` provisioning path) and one (missing `hack/helm/charts.yaml` in `helm-lint.yaml`'s path triggers) were addressed in this commit.

kpango:
Follow-up commit addressing feedback that Makefile.d/k8s.mk still hardcoded old-style flat chart names (`vald-helm-operator`, `vald-benchmark-operator`) inconsistent with the hierarchical `charts/operator/{helm,vald,benchmark}` layout established on `feature/vald-operator/initial-development`:

- Replaced the hardcoded chart-name path segments with a glob wildcard (`$(TEMP_DIR)/*/templates` → now per-recipe tmpdir, see below), so `k8s.mk` no longer depends on any chart's `Chart.yaml` `name:` field at all.
- A code-review pass caught that this wildcard was unsafe: `$(TEMP_DIR)` is a single directory memoized once per `make` invocation (not per-recipe), so under `make -jN` two recipes writing into it concurrently could race — in the worst case (`vald-operator` target) silently mixing another chart's YAML files into the wrong output. Fixed by giving each of the 7 affected recipes its own private `mktemp -d` directory instead of the shared one, chained into a single `set -e` + `trap ... EXIT`-guarded shell command so failures abort loudly and cleanup still happens.
- `k8s/manifest/all` switched from prerequisite-based aggregation to recursive `$(MAKE)` calls so its sub-targets can never run in parallel regardless of an outer `-jN`.
- Relocated `hack/helm/charts.yaml` → `charts/charts.yaml` (sits next to the charts it describes; confirmed no existing tool — kubeval/conflint, kube-linter, helm-docs — would misinterpret it), and dropped the now-redundant explicit path-trigger entry in `helm-lint.yaml` since it's covered by the existing `charts/**` glob.

Verified via `make -j3` running the three affected manifest-update targets concurrently (three genuinely distinct temp dirs, no cross-contamination), confirmed the `set -e`/`trap` pattern aborts correctly on failure while still cleaning up, and re-confirmed regenerated manifests byte-match committed baseline after the standard cleanup pass. Went through two more rounds of independent `code-reviewer` verification on this specific change; all raised issues (the TEMP_DIR race, the `k8s/manifest/all` parallel-aggregation gap, and the `;`-chain error-swallowing) are addressed in this commit.

kmrmt (approved):

datelier (approved):

kmrmt (approved):

datelier (approved):

